PR #24, opened 2019-02-12
Improve IPFS config

makeworld-the-better-one:
The main thing you missed is setting `Swarm.ConnMgr.Type` to `basic`.
> I tried to reduce this by setting the high and low water settings to a lower range, but this didn't seem to help.

This is because you didn't change the type, otherwise IPFS wouldn't do anything with those values.
> `Type` Sets the type of connection manager to use, options are: `"none"` (no connection management) and `"basic"`

So hopefully this will change things. I also set `Swarm.ConnMgr.GracePeriod` to `60s`, which means no connection will close until at least `60s` have passed, even if `HighWater` has been reached. This is recommended when lowering the `HighWater` and `LowWater` values, as it keeps connections from being closed abruptly. You might need to play with these values, but try this config out and see what you think.


makeworld-the-better-one:
I am confused as to why the Travis CI build failed. I am having trouble understanding the log. Does it matter?

sahib:
> I am confused as to why the Travis CI build failed. I am having trouble understanding the log. Does it matter?

My bad. Fix is in latest ``develop``.

sahib:
> So hopefully this will change things.

Cool! I'll give it a test.

makeworld-the-better-one:
@sahib Great! Let me know if it needs any changes, but everything should pretty much just work, I didn't change a lot. The testing is more long-term, about deciding on good values for `HighWater`, `LowWater` and `GracePeriod`.

sahib:
Out of curiosity I checked how the ``ConnMgr.Type`` is handled in IPFS:

It is read [here](https://github.com/ipfs/go-ipfs/blob/f749d5803cc02c78250ca65ebfc14c3b3f26073f/core/core.go#L311), which indicates that an empty string also selects basic but without picking up the low/high water settings. The default value seems to be already [basic](https://github.com/ipfs/go-ipfs-config/blob/master/init.go#L81) though. Also tested on my machine:

```bash
$ cat /tmp/ali/data/ipfs/config | jq '.Swarm.ConnMgr.Type' 
"basic"
```

I don't think setting it explicitly will make a difference, although it definitely won't hurt. I suggest we leave this PR open for now, if you find some other stuff to tune.

makeworld-the-better-one:
Interesting...

> if you find some other stuff to tune

I did change the `GracePeriod`, which may help. Also, I noticed you've enabled QUIC. Is it actually being used to connect to remotes?

sahib:
> Is it actually being used to connect to remotes?

I don't know. How can I tell? ``lsof`` would just show me open UDP connections. Probably the equivalent of ``ipfs swarm addrs``?

makeworld-the-better-one:
I just meant are you using the QUIC muiltaddr to connect to peers, such as `/ip4/<address>/udp/<port>/quic`?

sahib:
``brig`` dials to other peers [via their ID](https://github.com/sahib/brig/blob/master/net/client.go#L63), not by specifying an explicit address. Since I enabled QUIC, I assume that IPFS will use it if the remote peer is capable of it. But I'm not quire sure how to test that, since all of the code in ``brig`` is agnostic to the type of the connection - I only care if it is a more or less reliable stream.

makeworld-the-better-one:
To be honest I am unsure what IPFS will do at that level at all. Enabling QUIC won't hurt, but I have no idea whether it will use it there or not. Something to look into, perhaps.

sahib:
I will close that PR, since there are conflicts with ``develop`` due to the now externalized IPFS code. I added the grace period on my side in a82d3013.

Thanks for the efforts!

makeworld-the-better-one:
Ah okay, you're welcome, and thanks for adding it.

PR #59, opened 2020-12-24
Drastic simplification of IsCached

### LINKED ISSUE #54: Do not use heurestic of 262158 bytes to see if chunk does not have children
Currently there is a [heuristic](https://github.com/sahib/brig/blob/76c3335e7bec4c8643dcbc0ff061fda906ccb960/backend/httpipfs/pin.go#L212)
to see if a blob has children based on its size. It is used on defacto IPFS behavior but can change any moment.

We should disable it.

evgmik:
Closes #54

The old way was to manually recreate all children refs/links of a hash.
This required several calls to IPFS via api which was not efficient,
consequently we had 3 cache structures to track it.

Now,  all of this offloaded to IPFS which as I learned can do recursive
check for children of a hash/ref in one move. The speed seems to be
about the same, but we drastically simplified the code and
get rid of *two* unnecessary caches.

We could get rid of the last cache `locallyCached` as well but it still
gives some speed advantages on consequent calls for IsCached and cache
time is only 5 minutes. Nevertheless, it is something to consider.

The most likely scenario when we need real time updates:
a user pin a file which is remote and eagerly calls `brig ls` to see
if the status changed. But if it is big file it will probably take a few
minutes anyway, so it is OK to have a bit stale info.

### DIFF
--- a/backend/httpipfs/pin.go
+++ b/backend/httpipfs/pin.go
@@ -3,12 +3,9 @@ package httpipfs
 import (
 	"context"
 	"encoding/json"
-	"errors"
-	"fmt"
 	"io"
 	"strings"
 
-	"github.com/blang/semver"
 	"github.com/patrickmn/go-cache"
 	h "github.com/sahib/brig/util/hashlib"
 )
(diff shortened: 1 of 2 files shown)


sahib (approved):
Looks good to me. Less code is always good.

PR #60, opened 2020-12-24
Use StageFromReader instead of creation of temporary file

### LINKED ISSUE #58: stage --stdin operation does not remove temporary files
Subsequent runs `echo dummy | brig stage --stdin filename` leave undeleted buffer file in the `/tmp` directory. I.e. I see growing number of files like `/tmp/-brig.stdin.buffer4...`

evgmik:
This fixes #58. Old way: to stage from stdin a temporary file was created
but this file was not removed after use. Which polluted temporary
directories with such files.

New way: directly stage from stdin stream with StageFromReader.
Oddly, this function was already existing but was not used.

### DIFF
--- a/client/fs_cmds.go
+++ b/client/fs_cmds.go
@@ -146,6 +146,11 @@ func (cl *Client) StageFromReader(repoPath string, r io.Reader) error {
 	defer os.Remove(fd.Name())
 
 	if _, err := io.Copy(fd, r); err != nil {
+		fd.Close()
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


sahib (approved):
Approved, but I found a mistake of myself:

https://github.com/sahib/brig/blob/3c880455d299a46e6110c5032b0c2d367225831c/client/fs_cmds.go#L150

There should be a fd.Close() before the Stage() to not leak a file descriptor and make sure all content is flushed to the file.

PR #69, opened 2020-12-28
change encryption scheme (back) to random keys

### LINKED ISSUE #66: proposal: change encryption (back) to random key generation
On ``Stage()`` we're currently do the following things before adding the file to ``ipfs``, all of them requiring a seekable stream:

* We need the size of the stream. I build in a generic way that does not rely on files but on streams.
  Streams don't have a size and the only way to know is to either read all or seek to the end.
* The hash of the file is generated. This is used either as encryption key or to check if we need to do more work (if the content hash did not change, there's no need to stage the file).
* Part of the header is used to guess the compression algorithm we should use.

[See this function as reference](https://github.com/sahib/brig/blob/9f2e5a94c163eb67e5acc74540a11ceeaaed8f9b/catfs/fs.go#L870).

We probably can live without the compression guessing. Getting rid of the hash is harder, then we need another encryption guessing mechanism.  Current scheme is: If a file was initially added, compute the content hash+size and derive the key from it. If the file is modified the existing key is taken over. The idea here was that we want to enable de-duplication for same or similar files. This is however defeated by the way our encryption works: Each blocks influences the next block. So changes will result in a avalanche of changed cipher text. A single bit changed in the start of the file will therefore result in a completely different ciphertext. My plan to fix is to either wait for encryption support in IPFS or build some block layer in brig ourselves.

Until that happens (far future in any case) we should switch back to random key generation, which would allow us to get rid of the seekable stream requirement, also boosting performance since we need to read the stream only once.

### LINKED ISSUE #70: brig cat: big files are only outputted partially
**Describe the bug**

I added a big video (2G) via ``brig stage /file.mkv`` file and ``brig cat /file.mkv`` only outputted first 37 MB. Sometimes the size is also bigger. The syslog shows this line indicating a badly encoded stream or bad decoding:

```
28.12.2020/16:20:48 ⚠ server/fs_handler.go:172: IO failed for path /file.mkv on 127.0.0.1:46267: nonce size mismatch; should: 12 -have: 4
```

**To Reproduce**

See above.

**Expected behavior**

The full file to be outputted.

**Additional context**

Both IPFS and brig repositories are in /tmp.

sahib:
See #66


util/std.go
```diff
@@ -494,6 +494,43 @@ func PeekHeader(r io.ReadSeeker, size int64) ([]byte, io.ReadSeeker, error) {
...
+
+// Header returns the current header buffer.
+// It's size will be smaller or equal the size you passed to NewHeaderReader.
+func (hr *HeaderReader) Header() []byte {
+	return hr.buf
+}
+
+func (hr *HeaderReader) Read(buf []byte) (int, error) {
+	n, err := hr.r.Read(buf)
+	if diff := hr.size - uint64(len(hr.buf)); n > 0 && diff > 0 {
```
evgmik: Could this potentially backfire? `uint` minus another `uint`, might result in a negative number, which would appear humongous after casting to `uint`.  I remember fixing such problem in repiner.

evgmik (commented):
Looks like we lost our deduplication. We might search for old key based on Content Hash and reuse key and back end hash from this.

Other than my comment about `util/std.go`, changes seems to be good.

 The rest look good to me

catfs/fs.go
```diff
@@ -1021,6 +985,16 @@ func (fs *FS) Stage(path string, r io.ReadSeeker) error {
 		return err
 	}
 
+	// The stream was consumed, we now know those attrs:
+	size := sizeAcc.Size()
+	contentHash := hashWriter.Finalize()
+	compressAlgo, err = compress.GuessAlgorithm(path, headerReader.Header())
```
evgmik: We should run guessing before fs.bk.Add, here it useless and potentially incorrect.
By this time we already assigned default `comressAlgo` and used it with `mio` in `fs.bk.Add` couple lines above.

util/std.go
sahib: Oops, right. Luckily that case was caught by the inner if. Changed to a proper ``int64()`` here.

catfs/fs.go
sahib: Right, I had a logic error here. Now I remember why ``PeekHeader()`` actually did the reading before the add. :joy:
Seems I need to use the existing ``prefixReader`` but patch the part I read to the stream. Luckily Go makes that easy, that would be quite a horror in C/C++. Too bad I wrote some nice tests for ``HeaderReader``.

sahib:
> Looks like we lost our deduplication. We might search for old key based on Content Hash and reuse key and back end hash from this.

Might be an idea, but another partial fix. We use [AES-GCM](https://de.wikipedia.org/wiki/Galois/Counter_Mode) as cipher mode for encryption. This means effectively: one changed bit in the beginning yields a completely different cipher text anyways and kills deduplication. This needs a more general fix.

In an ideal world IPFs would do the encryption for us and we would not bother for the deduplication since IPFS would handle that for us. Alternatively we would have to do our own chunking or do some like git's [packfiles](https://git-scm.com/book/en/v2/Git-Internals-Packfiles) (i.e. re-encode several versions of a file to a highly compressed pack).

sahib:
By the way, you found two good points. :+1: Feel free to click "request changes" next time, not just comment. :smirk:

evgmik (approved):

sahib:
Side note: It seems that [upgrading FUSE breaks it](https://github.com/sahib/brig/pull/69/commits/d9514f4823c55967eb0b60085fed3715dd623744). I didn't really check what the cause was, but I think we should update since we're still using a version from 2018. Probably best to incrementally update and see where (and actually what) breaks.

@evgmik: You want to have a look on that?

evgmik:
> > Looks like we lost our deduplication. We might search for old key based on Content Hash and reuse key and back end hash from this.
> 
> Might be an idea, but another partial fix. We use [AES-GCM](https://de.wikipedia.org/wiki/Galois/Counter_Mode) as cipher mode for encryption. This means effectively: one changed bit in the beginning yields a completely different cipher text anyways and kills deduplication. This needs a more general fix.
> 
> In an ideal world IPFs would do the encryption for us and we would not bother for the deduplication since IPFS would handle that for us. Alternatively we would have to do our own chunking or do some like git's [packfiles](https://git-scm.com/book/en/v2/Git-Internals-Packfiles) (i.e. re-encode several versions of a file to a highly compressed pack).

I was thinking about a simplier case. What if we add files with different names but the same content (i.e. they are copies). In the old scheme, the key was derived from the content, so they end up with the same backend hash.  With random key, they endup with different backend hash but the same content hash. Since content is the same, we can check if we already have the same content hach, and use backend hash info from that node.

sahib:
> I was thinking about a simplier case. What if we add files with different names but the same content (i.e. they are copies). In the old scheme, the key was derived from the content, so they end up with the same backend hash. With random key, they endup with different backend hash but the same content hash. Since content is the same, we can check if we already have the same content hach, and use backend hash info from that node.

Yes, I understood that - that's why I called it a partial fix. It makes the situation a little better, but not much. Thus I don't consider it worthy to implement for now. The reason is that we use AES-GCM - changing a single bit early in the file will lead to a completely different text. Adding completely same content will result in deduplicated blocks (hopefully), but adding very similar files or modifying files will not.

Also there's a similar hen-egg problem as with the compression algorithm: to find out that we have the same content hash we need to read once through the stream (unless we do two passes as before). At that point the file is already encrypted.

evgmik (approved):

catfs/fs.go
```diff
@@ -955,35 +955,36 @@ func (fs *FS) Stage(path string, r io.Reader) error {
...
-			path,
-			headerBuf,
-		)
-
+		// Keep the header of the file in memory, so we can do some guessing
+		// of e.g. the compression algorithm we should use.
+		headerReader := util.NewHeaderReader(r, 2048)
+		headerBuf, err := headerReader.Peek()
 		if err != nil {
-			log.WithError(err).Warnf("failed to guess suitable zip algo for %s", path)
```
evgmik: Would failed attempt of `headerReader.Peek()` move the cursor position in `r`? Shall we reset it back to zero?

catfs/fs.go
```diff
@@ -955,35 +955,36 @@ func (fs *FS) Stage(path string, r io.Reader) error {
 		fs.cfg.String("compress.default_algo"),
 	)
 
+	// We might not be able to read the header for some reason.
```
evgmik: I would move the definition above the read of the default compression. Otherwise `err` check is visually disconnected from the code producing it.

fuse/file.go
```diff
@@ -3,12 +3,10 @@
 package fuse
 
 import (
+	"context"
 	"errors"
 	"os"
 	"sync"
-	"context"
-	"fmt"
```
evgmik, line 8: looks like linter went crazy. We need `fmt` and `log` for this module to compile

evgmik (changes requested):

fuse/file.go
sahib: No, I went crazy. Will fix once I found that bug.

catfs/fs.go
sahib: Done.

catfs/fs.go
sahib: Yes, it does move the cursor position, but headerReader preprends the buffer it read back to the stream.

sahib:
> If guess failed, `compressAlgo` potentially could be not set or damaged. We should add to the error handler:
> 
> ```
> compressAlgo = compress.AlgoNone
> ```
> 
> Since we use it in the couple places. We need to define `compressFallBack = compress.AlgoNone`

The zero value of ``compressAlgo`` is AlgoNone. So also when the variable would not be set, it would default to that.

evgmik (approved):

evgmik (approved):

evgmik:
Why the test commands are called ten-....?  I do not see logic of this name.

sahib:
> Why the test commands are called ten-....? I do not see logic of this name.

``ten`` because it prints (or validates) the numbers 0-9 (i.e. first ten). Liked it because it's short. Other suggestions?



Side note: I actually broke a PR best practice, because I got a little distracted during bug hunt. That PR got a little big and does more than what was initially advertised, should have done several smaller ones.

evgmik:
Regarding "ten". I understand now why it called that way. I see no other way to call it concisely.

PR #72, opened 2020-12-29
uncached files shown as broken symlinks in the fuse offline mode

### LINKED ISSUE #65: Proposal to show uncached files as simlinks with fuse
When we do fuse mount with `-o` (offline) option.
1. Let's assign file mode as symlink to uncashed files.
This way when someone asks for a listing, there will be nice visual (red font with colored list command) to remind that this file is uncached.

2. We might even show a link to the missing ipfs hash.

Item 2 is some extra work, but item 1 is easy. I already have a draft for it.

@sahib what do you think?

evgmik:
Closes #65

Additional/necessary changes

Added
func (fi *File) Readlink(ctx context.Context, req *fuse.ReadlinkRequest) (string, error)
to resolve name of the link.

Changed default write permissions
from `rwx rwx rwx` to `rw- r-- ---`

### DIFF
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -40,7 +40,6 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 	var filePerm os.FileMode = 0640
 	attr.Mode = filePerm 
 	if fi.m.options.Offline {
-		fmt.Printf("info %v\n", info)
 		isCached, err := fi.m.fs.IsCached(fi.path)
 		if err != nil || !isCached {
 			// uncached file will be shown as symlink
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"sync"
 	"context"
+	"fmt"
 
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
@@ -14,7 +15,7 @@ import (
 
 var (
 	// ErrNotCached is returned in offline mode when we don't have a file
-	ErrNotCached = errors.New("content is not cached and may not download")
+	ErrNotCached = errors.New("content is not cached and need to be downloaded")
 	ErrTooManyWriters = errors.New("too many writers for the file")
 )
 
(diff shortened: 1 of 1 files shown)


fuse/file.go
```diff
@@ -36,7 +37,18 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 	}
 	debugLog("exec file attr: %v", fi.path)
 
-	attr.Mode = 0755
+	var filePerm os.FileMode = 0640
+	attr.Mode = filePerm 
+	if fi.m.options.Offline {
+		fmt.Printf("info %v\n", info)
```
sahib: Please remove that debug log.

fuse/file.go
```diff
@@ -36,7 +37,18 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
...
+		isCached, err := fi.m.fs.IsCached(fi.path)
+		if err != nil || !isCached {
```
sahib, line 45: This will also show files as symlinks where ``IsCached`` errored for some reason. We should at least log the error.

fuse/file.go
```diff
@@ -168,6 +180,16 @@ func (fi *File) Listxattr(ctx context.Context, req *fuse.ListxattrRequest, resp
...
 }
 
+// Readlink reads a symbolic link.
+// This call is triggered when OS tries to see where symlink points
+func (fi *File) Readlink(ctx context.Context, req *fuse.ReadlinkRequest) (string, error) {
+	info, err := fi.m.fs.Stat(fi.path)
+	if err != nil {
+		return "/brig/backend/ipfs/", err
+	}
+	return fmt.Sprintf("/brig/backend/ipfs/%s", info.BackendHash), nil
```
sahib, line 193: Hmm, not so sure about the intention of this feature. What would the user supposedly do with the link in here?
Programs that call ``readlink()`` will detect this as ``ENOENT`` and just show a broken link. The text above would be only seen if somebody calls the ``readlink`` utility on the file.

sahib (changes requested):
Looks good to me overall, one clarification needed and two smaller nitpicks.

sahib:
It might be also worth to add a test for this behavior.

fuse/file.go
evgmik: Done with 3a5c46b

I have a question. The fuse files avoid using log facilities, i.e. `log.Debug()` and sparingly use `debugLog()`. Is there a deep reason for it? Is it used to enable/disable logging only for fuse module with `enableDebugLogs = false/true`  https://github.com/sahib/brig/blob/6140e52ffe5713c50f2ea01bb0cf9a0f0e69e799/fuse/fs.go#L15

fuse/file.go
evgmik: done with 7e125c5

fuse/file.go
sahib: No, I don't think there was a deeper reasons to it. FUSE debugging can get very spammy, so I wanted an easy flag I can toggle. Some day I will implement proper subsystem logging.

fuse/file.go
evgmik: If we do not implement `Readlink` then tools like `ls` would spam the output with
```
ls: cannot read symbolic link '/tmp/zz/README.md': Input/output error
```
Where do we point a link to noncached file is good question. I chose to point it quite wordy link which somewhat describe the issue. This way
a user can get some hints without calling `brig` (if someone is using fuse, they probably are shielded from `brig`). Also one can do a shell only script to ask `ipfs` to download missing hash.

sahib: Okay. I guess that's good enough for now. Maybe later we could do some more tricks with the link content.

sahib (approved):

PR #80, opened 2021-01-08
Fix sporadically failing fuse tests

### LINKED ISSUE #78: fix fuse tests
We observe that `fuse` test are occasionally fail. This was mentioned in several
followups to issues (for example #77).

As was summarized in that discussion the bug is actually in `go v1.9` and higher, which treats some polls as blocking. See discussion of this in
https://github.com/bazil/fuse/issues/264#issuecomment-727269770

On our sides it is broken because the same process manages `fuse` mount and do the file operations with it. Which occasionally triggers the infinite wait loop which freezes the test, and leaves a `defunct` test process and makes high `load` in linux.

Good news nothing is wrong with `fuse` layer, bad news we need to fix its tests.

`basil/fuse` suggest to run two separate fuse processes during tests.
https://github.com/bazil/fuse/commit/3f5ad9058d81bedd9d0181e676373e6b8053bb2a

evgmik:
Fixes #78
The fuse tests were sporadically failing often leaving <defunct>
processes which were loading process queue. I observed load of 30
for several our until reboot, though there we no negative effects.

After investigation, we found that this is due to `go` (v1.9 and up)
ways of handling poll requests. Here are the outcomes:

The routines which manage fuse layer
and OS dependent code (os.Open, and similar)
MUST BE SEPARATE OS PROCESSES.  Note: not different `go` routines but processes!
See https://github.com/bazil/fuse/issues/264#issuecomment-727269770
This separation happens automatically during normal brig operations, but
TESTING FUSE LAYER IN GO IS VERY TRICKY.
See brig relevant discussion at
https://github.com/sahib/brig/pull/77#issuecomment-754831080
However, this problem is general for any go program from version 1.9,
as can be seen in references to the issue.

bazil/fuse offers "bazil.org/fuse/fs/fstestutil/spawntest"
infrastructure which helps run tests in different communicating via socket processes.

This commit rewrites fuse test to use methods from `spawntest`.

### DIFF
--- a/fuse/fuse_test.go
+++ b/fuse/fuse_test.go
@@ -49,60 +49,60 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-type brigHelp struct{}
+type fuseCatFSHelp struct{}
 
 // These helpers will be requested from test and executed on the server
-// which is managing brig-fuse connection (started within test)
-func (bmh *brigHelp) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+// which is managing catfs-fuse connection (started within test)
+func (fch *fuseCatFSHelp) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	switch req.URL.Path {
 	case "/mount":
-		httpjson.ServePOST(bmh.makeBrigAndFuseMount).ServeHTTP(w, req)
+		httpjson.ServePOST(fch.makeCatfsAndFuseMount).ServeHTTP(w, req)
 	case "/unmount":
-		httpjson.ServePOST(bmh.unmountFuseAndcloseDummyBrigFS).ServeHTTP(w, req)
+		httpjson.ServePOST(fch.unmountFuseAndCloseDummyCatFS).ServeHTTP(w, req)
 	case "/fuseReMount":
-		httpjson.ServePOST(bmh.makeFuseReMount).ServeHTTP(w, req)
-	case "/brigStage":
-		httpjson.ServePOST(bmh.brigStage).ServeHTTP(w, req)
-	case "/brigGetData":
-		httpjson.ServePOST(bmh.brigGetData).ServeHTTP(w, req)
+		httpjson.ServePOST(fch.makeFuseReMount).ServeHTTP(w, req)
+	case "/catfsStage":
+		httpjson.ServePOST(fch.catfsStage).ServeHTTP(w, req)
+	case "/catfsGetData":
+		httpjson.ServePOST(fch.catfsGetData).ServeHTTP(w, req)
 	default:
 		http.NotFound(w, req)
 	}
 }
 
-func makeDummyBrigFS(dbPath string) (brigFuseInfo, error) {
+func makeDummyCatFS(dbPath string) (catfsFuseInfo, error) {
 	backend := catfs.NewMemFsBackend()
 	owner := "alice"
 
 	cfg, err := config.Open(nil, defaults.Defaults, config.StrictnessPanic)
 	if err != nil {
-		log.Fatalf("Could not get default brig FS config: %v", err)
-		return brigFuseInfo{}, err
+		log.Fatalf("Could not get default catFS config: %v", err)
+		return catfsFuseInfo{}, err
 	}
 
-	bfs, err := catfs.NewFilesystem(backend, dbPath, owner, false, cfg.Section("fs"))
+	cfs, err := catfs.NewFilesystem(backend, dbPath, owner, false, cfg.Section("fs"))
 	if err != nil {
-		log.Fatalf("Failed to create brig filesystem: %v", err)
-		return brigFuseInfo{}, err
+		log.Fatalf("Failed to create catfs filesystem: %v", err)
+		return catfsFuseInfo{}, err
 	}
-	bfInfo := brigFuseInfo{}
-	bfInfo.bfs = bfs
-	bfInfo.dbPath = dbPath
+	cfInfo := catfsFuseInfo{}
+	cfInfo.cfs = cfs
+	cfInfo.dbPath = dbPath
 
-	return bfInfo, err
+	return cfInfo, err
 }
 
 type nothing struct{} // use it to send empty request or responses to server
 
-type brigFuseInfo struct {
-	bfs       *catfs.FS
+type catfsFuseInfo struct {
+	cfs       *catfs.FS
 	dbPath    string
 	fuseMount *Mount
 }
 
-// bfInfo will be in the global space for the server
-// which manage fuse mount connection to the brig FS
-var bfInfo brigFuseInfo
+// cfInfo will be in the global space for the server
+// which manage fuse mount connection to the catFS
+var cfInfo catfsFuseInfo
 
 type mountingRequest struct {
 	DbPath  string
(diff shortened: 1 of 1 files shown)


fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+type mountInfo struct { // fuse related info available to OS layer
+	Dir  string
+	Opts MountOptions
 }
 
-func checkForCorrectFile(t *testing.T, path string, data []byte) {
-	// Try to read it over fuse:
-	helloBuffer := &bytes.Buffer{}
-	fd, err := os.Open(path)
+func withMount(t *testing.T, opts MountOptions, f func(ctx context.Context, control *spawntest.Control, mount *mountInfo)) {
```
sahib, line 231: I saw that you often pass the zero value of MountOptions to ``withMount()``. It's common practice to make ``opts`` a pointer and fill the defaults in ``withMounts`` if it's nil. Just a small nitpick, but makes code a bit tidier I think.

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+
+	return &out, err
 }
 
-func withMount(t *testing.T, opts MountOptions, f func(mount *Mount)) {
-	withDummyFS(t, func(fs *catfs.FS) {
-		withMountFromFs(t, opts, fs, f)
-	})
+var helpers spawntest.Registry
+var brigHelper = helpers.Register("brigHelp", &brigHelp{})
```
sahib: Prefixing names with the program name is rather unusual in Go, I suggest to use more specific names like ``fuseHelper`` or something.

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+
+	control := brigHelper.Spawn(ctx, t)
+	defer control.Close()
+
+	dbPath, err := ioutil.TempDir("", "brig-fs-test")
 	if err != nil {
-		t.Fatalf("Unable to open simple file over fuse: %v", err)
+		t.Fatalf("Failed to create temp dir for brig file system: %v", err)
 	}
+	require.Nil(t, err)
```
sahib: Error double check?

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+		Opts:    opts,
+	}
 
-	n, err := io.CopyBuffer(helloBuffer, fd, make([]byte, 128*1024))
-	if err != nil {
-		t.Fatalf("Unable to read full simple file over fuse: %v", err)
+	if err := control.JSON("/mount").Call(ctx, req, &nothing{}); err != nil {
+		t.Fatalf("calling helper: %v", err)
 	}
+	require.Nil(t, err)
```
sahib: Error double check?

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
 
-	if n != int64(len(data)) {
-		t.Fatalf("Data differs over fuse: got %d, should be %d bytes", n, len(data))
+	mount := mountInfo{Dir: req.MntPath, Opts: req.Opts}
+	// function which required mounts
+	f(ctx, control, &mount)
+
+	// cleanup and unmount
+	if err := control.JSON("/unmount").Call(ctx, nothing{}, &nothing{}); err != nil {
```
sahib: It might be an idea to execute this code in a defer func, so it gets executed always, even if the test panics. Otherwise we might still have a (dysfunctional) mount lying around in that case.

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+	require.Nil(t, err)
+}
 
-	if !bytes.Equal(helloBuffer.Bytes(), data) {
-		t.Errorf(
-			"Data from simple file does not match source. Len: %d %d",
-			len(data),
-			helloBuffer.Len(),
-		)
+func checkFuseFileMatcheToBrig(t *testing.T, ctx context.Context, control *spawntest.Control, fusePath string, brigPath string) {
```
sahib: Small typo: ``Matche``.

fuse/fuse_test.go
```diff
@@ -17,104 +19,290 @@ import (
...
+	if err := control.JSON("/mount").Call(ctx, req, &nothing{}); err != nil {
+		t.Fatalf("calling helper: %v", err)
 	}
+	require.Nil(t, err)
 
-	if n != int64(len(data)) {
-		t.Fatalf("Data differs over fuse: got %d, should be %d bytes", n, len(data))
+	mount := mountInfo{Dir: req.MntPath, Opts: req.Opts}
+	// function which required mounts
+	f(ctx, control, &mount)
```
sahib: More a FYI: When I use a variable only to pass it to a function, I usually pass it as literal, i.e:

```go
f(ctx, control, &mountInfo{
    Dir: req.MntPath,
    Opts: req.Opts
})
```

Just reduces my mental overhead when reading the code, since I can be sure that the variable is not used anywhere else and I don't need to check.

sahib (changes requested):
Good work! All comments are more about style, less about functionality.

fuse/fuse_test.go
evgmik: My bad: `brigFS` is officially `catfs`. But I did this mental substition through the `fuse-test` code.  Why do you call it `cat...`, is it short for catalog?

But I see why it would be confusing, for those who read the code. I will relabel the problem spots.

sahib: It stands for **c**ontent **a**ddressed **t**ransactional **f**ile**s**ystem. But that's just alibi, I just like cats.

fuse/fuse_test.go
evgmik: I would like to keep it this way. Otherwise I had to constantly check the definition of the function to see what place the nil is taking.

My way is good for copy-pasting a test header, I see exactly what is sent to withMount. It is more typing if you do it from scratch, but who does it?

Am I missing something?

sahib: I'm not opposed to leaving it as it is. I just took it as an example for a typical Go-technique. So I will approve both ways - it's more of a philosophical question after all.

> Otherwise I had to constantly check the definition of the function to see what place the nil is taking.

Fair point, but usually good text editors IDEs help with that. If I don't remember what e.g. the third parameter I just jump to the function definition. If I always have to do it, it's usually a sign for me to simplify the function because it might have too many parameters. Arguably languages like python make this easier by the optional use of keyword arguments like ``f(buffer_size=4)``. 

> ...but who does it?

Me often, when I'm unfamiliar with some code, I try to avoid copy&pasting. Typing function calls makes you think. Later on I do plenty of copy&paste too, though. Anyway, APIs should be designed to be easy to be used and understood, not to be easy to copy & paste.

sahib (approved):

fuse/fuse_test.go
evgmik: fixed in 5a940b7

fuse/fuse_test.go
evgmik: fixed in 5a940b7

fuse/fuse_test.go
evgmik: done in 34b7988

fuse/fuse_test.go
evgmik: fixed with 3962a45

fuse/fuse_test.go
evgmik: fixed in bb37469

fuse/fuse_test.go
evgmik: addressed with massive renaming in f33b437

fuse/fuse_test.go
evgmik: I left it as it were. But it is time for me to tune up my Vim

PR #81, opened 2021-01-10
Fix/build adjustment for obsolete toolchains

evgmik:
This will help to build `brig` to those of us who runs obsolete (stable) tool chains. Here I refer to Debian stable (buster) which has quite stale git and capnp.

Changes:
- build works with older version of git
- if capnp and similar resources do not need to be regenerated (cloned directly from our repo), skip `generate` step

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -38,7 +38,6 @@ tasks:
       - ./gateway/static/**/**/**/**
 
   build:
-    deps: [generate]
     desc: "Build the brig binary"
     cmds:
       - ./scripts/build.sh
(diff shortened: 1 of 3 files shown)


Taskfile.yml
```diff
@@ -48,6 +47,10 @@ tasks:
       - ./*.go
       - ./**/*.go
 
+  dev-build:
+    deps: [generate, build]
+    desc: "Developer build from scratch (regenerates internal resources)"
+
```
sahib: I think the better fix would be to check in ``scripts/generate.sh`` to check if the ``capnp`` executable is available. No need for an additional build target then.

evgmik: Did as you suggested in a118c47 and reverted Taskfile back with  27189de

sahib (approved):

Issue #68, opened 2020-12-28
provide automated benchmarks for streaming layers


sahib:
``brig`` has many layers of streaming and for regression and performance debugging, we should provide automated benchmarks.
The following things should be measured in throughputs:

* Throughput of our internal ``ipfs cat`` and ``ipfs add`` implementations.
* Throughput of the compression/decompression layer.
* Throughput of the encryption/decryption layer.
* Throughput of TCP socket for the ``brig cat`` case.
* Throughput of File reading from /tmp for the ``brig stage`` case.
* Throughput of a simple copy from stdin to stdout <sup>1</sup>

<sup>1</sup> this is the I/O baseline your CPU can do - that's roughly 32 MB/s for my rather slow laptop.

sahib:
Extensively covered by #93.

PR #82, opened 2021-01-11
Enhancement/benchmarks fuse

evgmik:
Fuse read and write speed benchmarks. Works towards #68


fuse/fuse_test.go
```diff
@@ -480,3 +480,99 @@ func TestWithRoot(t *testing.T) {
...
+			catfsFilePath := fmt.Sprintf("%s_file_%d", label, size)
+			fuseFilePath := filepath.Join(mount.Dir, catfsFilePath)
+
+			b.StartTimer()
+			for i := 0; i < ntimes; i++ {
+				if len(data)>0 {
+					// generate new content for backend
+					data[0]++
+				}
+			}
```
sahib, line 551: ``ntimes`` only controls how often ``data`` is touched, not how often we actually write. Is that on purpose? Also the ``data`` modification is counted as the benchmark time, which I think it should not.

fuse/fuse_test.go
```diff
@@ -480,3 +480,99 @@ func TestWithRoot(t *testing.T) {
...
+		data := testutil.CreateDummyBuf(size)
+		prefix := fmt.Sprintf("Owerwrite_%d", Ntimes)
+		label := fmt.Sprintf("%s/CompressibleContent_Size_%d", prefix, size)
+		b.Run(label, func(b *testing.B) {
+			writeDataNtimes(b, data, Ntimes)
+		})
+
+		// Check how fast is write to a file with random/uncompressible content
+		data = testutil.CreateRandomDummyBuf(size, 1)
+		label = fmt.Sprintf("%s/RandomContent_Size_%d", prefix, size)
```
sahib, line 578: :+1: for having that as extra case vs easily compressible.

fuse/fuse_test.go
```diff
@@ -480,3 +480,99 @@ func TestWithRoot(t *testing.T) {
...
+		withMount(b, MountOptions{}, func(ctx context.Context, control *spawntest.Control, mount *mountInfo) {
+			// Check how fast is write to a file with compressible content
+			catfsFilePath := fmt.Sprintf("%s_file_%d", label, size)
+			fuseFilePath := filepath.Join(mount.Dir, catfsFilePath)
+
+			b.StartTimer()
+			for i := 0; i < ntimes; i++ {
+				if len(data)>0 {
+					// generate new content for backend
+					data[0]++
```
sahib: Really minor nitpick: If ``ntimes`` would go over 256 easily, you could also use ``binary.LittleEndian.PutUint64()``  to write the full number to the first few bytes.

fuse/fuse_test.go
```diff
@@ -480,3 +480,99 @@ func TestWithRoot(t *testing.T) {
...
+		1*1024*1024, 16*1024*1024,
+	}
+)
+
+func stageAndRead(b *testing.B, ctx context.Context, control *spawntest.Control, mount *mountInfo, label string, data []byte) {
+	size := len(data)
+	// stage data to catFS
+	catfsFilePath := fmt.Sprintf("%s_file_%d", label, size)
+	req := catfsPayload{Path: catfsFilePath, Data: data}
+	require.Nil(b, control.JSON("/catfsStage").Call(ctx, req, &nothing{}))
```
sahib: I nowadays recommend to use ``require.NoError`` since it produces better error messages when something fails. There's many places where around ``brig`` where still ``require.Nil`` is used, so I won't mind if you change them when you modify something around there.

sahib (commented):
Good first step.

Go's benchmark is a bit more targeted at CPU (unit: seconds) or memory (unit: bytes allocated / number of allocations) intensive workloads and less to I/O (unit: throughput as bytes per second). While you can more or less convert the results of the test to throughput (number of bytes are hardcoded in the code) it would be useful to directly display the throughput a certain benchmark has.

Maybe this could be also done outside the ``fuse`` package, since it will later include also benchmarks for things like ``brig cat`` and ``brig stage`` or more isolated benchmarks like encryption or compression benchmarks? 

Also maybe this could be instead implemented as ``brig`` command? Something like ``brig debug iobench`` which users could easily execute as part of bug reports. Just leaving that as idea to discuss.

evgmik:
> Good first step.

I would call it half a step, since it is tested with "fake" `catfs` without reflecting to `ipfs`. But I am about to start working on replacement `fuse` holding files in memory during modification, so I needed at least some benchmarks.


> 
> Maybe this could be also done outside the `fuse` package, since it will later include also benchmarks for things like `brig cat` and `brig stage` or more isolated benchmarks like encryption or compression benchmarks?
> 
> Also maybe this could be instead implemented as `brig` command? Something like `brig debug iobench` which users could easily execute as part of bug reports. Just leaving that as idea to discuss.

Cleaning up after such test might be tricky, but I like the idea.

fuse/fuse_test.go
evgmik: Replaced accordingly  everywhere in fuse-test.go with 0fa89b70

I would say we will need similar global patch, but this is for another branch.

fuse/fuse_test.go
evgmik: Random content is about factor of 3 slower to read and write. Not sure what to blame. Is it because backend stores or serve less data? Why is it factor of 3?

sahib: I guess it's because it's way more data that needs to be encrypted / decrypted. But that's the reason why we need also benchmarks for parts of the pipeline.

fuse/fuse_test.go
sahib: Either that or just replace it along other modifications.

sahib:
> I would call it half a step, since it is tested with "fake" catfs without reflecting to ipfs. But I am about to start working on replacement fuse holding files in memory during modification, so I needed at least some benchmarks.

Well, that's not so bad. ``catfs`` is just in memory and will probably output more stable numbers than what ipfs with it's many I/O calls will do. Sure, those are not the numbers one will actually get when using the fuse layer with ipfs, but they should be easier to compare.

> Cleaning up after such test might be tricky, but I like the idea.

Why would it be tricky?

----

If we go for ``brig debug iobench`` we could also output various system information that influences the benchmark results (CPU model, IPFS version etc.). All throughput numbers can also be expressed additionally as relative to the baseline (e.g. just copying bytes from ``ten-source`` to ``ten-sink``).

fuse/fuse_test.go
evgmik: Not sure that I understand. Why would `ntimes` go over 256. It is a constant in this code.

As for stamping the data. It content really is not important, as long as, at least one byte changes to produce a new `data` hash.

fuse/fuse_test.go
evgmik: nice catch with ntimes. Shame on me. Fixed with a4001070

 As for data modification, the idea
is to flip only one the very first byte (which instantaneous), with idea that encryption would scramble it to totally new content.

Note to myself, I need to write  fuse Seek implementation.

evgmik:
> Well, that's not so bad. `catfs` is just in memory and will probably output more stable numbers than what ipfs with it's many I/O calls will do. Sure, those are not the numbers one will actually get when using the fuse layer with ipfs, but they should be easier to compare.

There is a part of in memory which sits on hard drive. I am guessing it is the badger or something. When we mount `catfs` we provide a path for it; its content size grows very quicklu with every write. Makes me wonder if it is truly "in memory".

> > Cleaning up after such test might be tricky, but I like the idea.
> 
> Why would it be tricky?
If we do test on a live brig instance, we would have to write quite a lot to `ipfs` and to the `catfs`. To clean up, we have to be track what was additionally written and remove such haches from `ipfs` (with potential for real content overlap and removal of something important),also `catfs` had to be reverted to the before benchmark state. We can do it via commits, but it would contaminate history with "benchmark" commits.

Another issue, what if a user modifies something during benchmarks. How to separate it?


> If we go for `brig debug iobench` we could also output various system information that influences the benchmark results (CPU model, IPFS version etc.). All throughput numbers can also be expressed additionally as relative to the baseline (e.g. just copying bytes from `ten-source` to `ten-sink`).

I am setting it as v0.7 target.

fuse/fuse_test.go
sahib: > Not sure that I understand. Why would ntimes go over 256. It is a constant in this code.

It does not in the current code, but always imagine someone is changing the code that doesn't know it yet.
This someone could easily decide to add it to ``NumberOfOverWrites`` and expect that everything works.
It's just defensive programming and making code future proof for changes. After all ``ntimes`` is passed as ``int``.
> As for stamping the data. It content really is not important, as long as, at least one byte changes to produce a new data hash.

Yep, got that. That's why I got curious why only the first byte was written and what the edge cases are.

fuse/fuse_test.go
sahib: > is to flip only one the very first byte (which instantaneous), with idea that encryption would scramble it to totally new content.

True, the encryption will cause totally different blocks for different start bytes.

> Note to myself, I need to write fuse Seek implementation.

What do you mean?

sahib (commented):
> There is a part of in memory which sits on hard drive. I am guessing it is the badger or something. When we mount catfs we provide a path for it; its content size grows very quicklu with every write. Makes me wonder if it is truly "in memory".

Hmm, interesting. ``catfs.MemFsBackend`` actually is just memory - you can check the implementation in ``catfs/backend.go``.
The metadata database however is indeed a badger database. That might indicate that many writes fill it up quickly. Can you check what those temporary files are?

> If we do test on a live brig instance [...]

No need to do it on a live instance. For this purpose we can easily add a dummy ``ipfs`` instance that has it's repo in ``/tmp`` and is only used for our benchmarking and will be deleted after. See ``scripts/test-bed.sh`` on how to do it. We also have Go code to initialize a fresh ``ipfs`` repo. For the brig server part we can just spin up a new server in the benchmark process itself with ``server.BootServer()``. 

> I am setting it as v0.7 target.

Ok.

evgmik:
> > There is a part of in memory which sits on hard drive. I am guessing it is the badger or something. When we mount catfs we provide a path for it; its content size grows very quicklu with every write. Makes me wonder if it is truly "in memory".
> 
> Hmm, interesting. `catfs.MemFsBackend` actually is just memory - you can check the implementation in `catfs/backend.go`.
> The metadata database however is indeed a badger database. That might indicate that many writes fill it up quickly. Can you check what those temporary files are?

`NewFileSystem` creates badger db with `db.NewBadgerDatabase(dbPath)` even if `catfs` is in memory. Many (thousands )consequent writes (which Benchmarks like to do), grow this DB to quite large size. This is why I had to unmount/remount `catfs` in write benchmarks.

evgmik:
> >    Note to myself, I need to write fuse Seek implementation.

> What do you mean?

It is me being silly. I forget that read/write sets the cursor, without prior call to Seek.

fuse/fuse_test.go
evgmik: > > Not sure that I understand. Why would ntimes go over 256. It is a constant in this code.
> 
> It does not in the current code, but always imagine someone is changing the code that doesn't know it yet.
> This someone could easily decide to add it to `NumberOfOverWrites` and expect that everything works.
> It's just defensive programming and making code future proof for changes. After all `ntimes` is passed as `int`.

Now I am confused, why we are worried about `256`? Neither `i` nor `ntimes` are used inside the loop
```
for i := 0; i < ntimes; i++ {
	if len(data) > 0 {
		// modification of one byte is enough
		// to generate new encrypted content for the backend
		data[0]++
	}
	require.NoError(b, ioutil.WriteFile(fuseFilePath, data, 0644))
}
```
`i:=0` is untyped but defaults to `int` (32 bits by go standard) which is the same as `ntimes`, so there should be no overflow. Unless some one puts a negative number for `ntimes` we should be OK. 

I added protection against negative input in ae4ba980

sahib: It's not used inside the loop, but you increment ``data[0]`` ``ntimes``. If ``ntimes > 255`` the byte overflows and you will get a hash that you had before. This will be (might be?) faster than truly unique data and could influence the benchmark results. Therefore my comment about encoding ``i`` completely in the first few bytes with ``binary.LittleEndian.PutUint64()``.

Sure, it's a little contrived example considering that this won't happen in the test (``ntimes`` is always low), I just saw a opportunity for some general advice. Hope you don't mind.

evgmik: Now I get it. You are precisely correct and I am looking at wrong place.
Fixed with 65382764

sahib (approved):

PR #84, opened 2021-01-13
Massive redesign to avoid use of MaxUint64 as indicator of unknown size

### LINKED ISSUE #53: Do not use MaxUint64 as indicator of unknow file size.
This is technically not a bug but make trigger some if a user
would use output of 
[CachedSize](
https://github.com/sahib/brig/blob/76c3335e7bec4c8643dcbc0ff061fda906ccb960/backend/httpipfs/pin.go#L237
) in something like `make([]byte, size)`.

What should be done:
Size should use int64 instead uint64, negative values should reinforce *unknown size* error/warning.

evgmik:
Addresses #53
Not sure I like the changes, quite a lot of utilities assumed and worked
with uint64 integers. But this is "no go", if we want to use negative size
as indicator of an error or "do not use me" indicator.

Looks like we are reducing size capacity by 2 (we are not using the sign bit).
I understand that if we using 2^63 file size, then we are in far future,
but I fill guilty by putting factor of 2 reduction.

What I like the least, I had replace uint64 with int64 in mio/stream

But if we OK with this, then changes pass all tests and ready to go.


sahib:
I'm afraid we had a misunderstanding here. I was referring to making ``CachedSize`` a ``int64`` since it might have a *"we don't know"* state. That does not apply to ``Size`` though. We always have a size and the type of ``uint64`` indicates that. That should massively cut down the number of lines you have to change.

> Looks like we are reducing size capacity by 2 (we are not using the sign bit).
> I understand that if we using 2^63 file size, then we are in far future,
> but I fill guilty by putting factor of 2 reduction.

If somebody tries to store 8192 Petabyte in brig he needs to bring a lot of patience anyways. If he comes that far.

evgmik:
> I'm afraid we had a misunderstanding here. I was referring to making `CachedSize` a `int64` since it might have a _"we don't know"_ state. That does not apply to `Size` though. We always have a size and the type of `uint64` indicates that. That should massively cut down the number of lines you have to change.

No worries, I should have clarified ("everything, which could be misinterpreted, will be" :).

What would be the best way to send only `CachedSize` related modifications? New branch or reversion of the commit in this one?

sahib:
> What would be the best way to send only CachedSize related modifications? New branch or reversion of the commit in this one?

A ``git revert`` of this commit is fine. Or just change the parts back. The diff to ``develop`` is what matters in the end. :smile:

evgmik:
> > What would be the best way to send only CachedSize related modifications? New branch or reversion of the commit in this one?
> 
> A `git revert` of this commit is fine. Or just change the parts back. The diff to `develop` is what matters in the end. smile

Ok. I did revert with 9391e7d and did only relevant to `CacheSize` fixes in 987cf49

sahib (approved):

sahib:
Looks good now. :+1:

PR #86, opened 2021-01-16
Refactor/repo layout

sahib:
So, this is still work in progress and might change a little.

In summary the changes are:

* All password handling has been stripped from the daemon. The advantage is that we do not need to handle passwords anymore or have code to implement things like ``passwd``. 
* The repository layout changed in a backwards-incompatible way in an attempt to clean things up. Since we don't have to decide anymore between "files we need to start the daemon that can unlock the repo" and "files that should be encrypted when locking". we can just tar and encrypt the whole directory. 
* The new ``pack-repo`` and ``unpack-repo`` command can be used to convert a repository in an encrypted tar archive and read it back. This is meant as replacement of the existing, but messy locking code we had now. In contrast, it is supposed to be called and hooked up by the user. On request we can think of having a config option to unpack/pack a repository as a best attempt upon daemon start/shutdown.

TODO:

* ~~Adjust tests, if any are broken.~~
* ~~See if we can delete more code.~~
* ~~Make nomenclature consistent.~~
* ~~Add a "is-packed" command for debugging.~~ Not needed.
* ~~General clean up.~~
* ~~Update docs.~~


evgmik (approved):

evgmik:
Shall we abandon all password locking? If repo is locked for other users with no read/write permission, then nobody but a user can access it anyway. If one wants to move it another computer, then the copy either protected by the same means or should be send in an encrypted way (locked archive). But there is no need to build such capability in brig. Password protected `zip` or `scp`  are OS provided solutions.

sahib:
Good point. I would still make the pack commands available as an experiment. My argument is that many people might no even be aware that it's a good idea to move the repository in an encrypted way. Offering a command that does the "right" thing (including checking for running daemon & fixing the permissions if needed) lowers the barrier to do it. Sometimes you might not have access to the proper tools or are working in an environment where other people work, e.g. imagine you live together with some room mates that use a common file server, then you want to make sure that after every use the brig repository is locked - sure, that's not the normal way to use our tool, but the first thing people do is to misuse it anyways. Also, it's not a lot of code & if it turns out that nobody uses it, I'll happily remove it.

Reminds me to check that the repo is created with the right permissions. Anyways, I don't want to put a lot of work into it now.

evgmik (approved):

evgmik (approved):

PR #88, opened 2021-01-23
Feature/fuse improvements

### LINKED ISSUE #49: fix: stabilize & improve the FUSE layer
Things that should be tackled:

* Current implementation holds a buffer per file (if opened rw) with all its contents.
* This is memory intensive, so the old state (``catfs/mio/overlay``) should be partly reverted.
* Overlay system should be improved so we don't hold all data in memory when writing a large, completely new file.
* Reading small files can be optimized by reading all of it and directly in the overlay layer. If [layer.Read()](https://github.com/sahib/brig/blob/develop/catfs/mio/overlay/overlay.go#L248) does not need to go to the underlying stream it will be a lot faster.

(Note: The above could be implemented by using BadgerDB instead of [keeping modifications in memory](https://github.com/sahib/brig/blob/develop/catfs/mio/overlay/overlay.go#L23))

Other things:

* Check for race conditions in the FUSE layer.

evgmik:
Adresses #49 
Enabling `fuse` optimizations at mount: Cache and larger buffer sizes. This gives quite large speed improvements, I even had to think about disabling  debug outputs, since it takes comparable time to disk io.

Also, removed use of write buffer back to `overlay` system. 

Main slow down for using `overlay` was useless `Seek` at write: we write to memory overlay but aks the backend to `Seek` for no reason.  This problem solved with `WriteAt` method.

@sahib please pay attention how the `overlay` limits are handled. I am not sure that I fully grasped what for they are used. The structure field had no comment :(

### DIFF
--- a/catfs/mio/overlay/overlay.go
+++ b/catfs/mio/overlay/overlay.go
@@ -342,6 +342,11 @@ func (l *Layer) Seek(offset int64, whence int) (int64, error) {
 		l.limit = newPos
 	}
 
+	if l.pos == newPos {
+		// very likely it sequent read/write request
+		// no need to seek since we already pointing to the right position
+		return l.pos, nil
+	}
 	l.pos = newPos
 
 	// Silence EOF:
--- a/catfs/mio/overlay/overlay.go
+++ b/catfs/mio/overlay/overlay.go
@@ -228,6 +228,24 @@ func (l *Layer) Write(buf []byte) (int, error) {
 	return len(buf), nil
 }
 
+// Writes data from `buf` at offset `off` counted from the start (0 offset).
+// Mimics `WriteAt` from `io` package https://golang.org/pkg/io/#WriterAt
+func (l *Layer) WriteAt(buf []byte, off int64) (n int, err error) {
+	// Copy the buffer, since we cannot rely on it being valid forever.
+	modBuf := make([]byte, len(buf))
+	copy(modBuf, buf)
+
+	l.index.Add(&Modification{off, modBuf})
+	end := off + int64(len(buf))
+	if l.limit >= 0 && end > l.limit {
+		l.limit = end
+	}
+
+	n = len(buf)
+	err = nil
+	return n, nil
+}
+
 // hasGaps checks if overlays occludes all bytes between `start` and `end`
 func hasGaps(overlays []Interval, start, end int64) bool {
 	diff := end - start
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -56,11 +56,7 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 			attr.Mode = os.ModeSymlink | filePerm
 		}
 	}
-	if fi.hd != nil && fi.hd.writers > 0 {
-		attr.Size = uint64(len(fi.hd.data))
-	} else {
-		attr.Size = info.Size
-	}
+	attr.Size = info.Size
 	attr.Mtime = info.ModTime
 	attr.Inode = info.Inode
 
@@ -105,31 +101,12 @@ func (fi *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.Open
 
 	fi.mu.Lock()
 	if fi.hd == nil {
-		hd := Handle{fd: fd, m: fi.m, writers: 0, wasModified: false, currentFileReadOffset: -1}
+		hd := Handle{fd: fd, m: fi.m, wasModified: false, currentFileReadOffset: -1}
 		fi.hd = &hd
 	}
 	fi.hd.fd = fd
 	fi.mu.Unlock()
-	if req.Flags.IsReadOnly() {
-		// we don't need to track read-only handles
-		// and no need to set handle `data`
-		resp.Flags |= fuse.OpenKeepCache
-		return fi.hd, nil
-	}
 
-	// for writers we need to copy file data to the handle `data`
-	fi.hd.mu.Lock()
-	defer fi.hd.mu.Unlock()
-	if fi.hd.writers == 0 {
-		err = fi.hd.loadData(fi.path)
-		if err != nil {
-			return nil, errorize("file-open-loadData", err)
-		}
-	}
-	if fi.hd.writers == (^uint(0)) { // checks against writers overflow
-		return nil, errorize(fi.path, ErrTooManyWriters)
-	}
-	fi.hd.writers++
 	resp.Flags |= fuse.OpenKeepCache
 	return fi.hd, nil
 }
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -164,7 +164,9 @@ func (fi *File) Listxattr(ctx context.Context, req *fuse.ListxattrRequest, resp 
 	defer logPanic("file: listxattr")
 	log.Debugf("fuse-file-listxattr: %v", fi.path)
 
-	resp.Xattr = listXattr(req.Size)
+	// Do not worry about req.Size
+	// fuse will cut it to allowed size and report to the caller that buffer need to be larger
+	resp.Xattr = listXattr()
 	return nil
 }
 
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -165,7 +165,7 @@ func (fi *File) Fsync(ctx context.Context, req *fuse.FsyncRequest) error {
 
 // Getxattr is called to get a single xattr (extended attribute) of a file.
 func (fi *File) Getxattr(ctx context.Context, req *fuse.GetxattrRequest, resp *fuse.GetxattrResponse) error {
-	// defer logPanic("file: getxattr")
+	defer logPanic("file: getxattr")
 	// log.Debugf("fuse-file-getxattr %v for atribute %v", fi.path, req.Name)
 
 	// debugLog("exec file getxattr: %v: %v", fi.path, req.Name)
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -150,7 +150,10 @@ func (fi *File) Getxattr(ctx context.Context, req *fuse.GetxattrRequest, resp *f
 	// log.Debugf("fuse-file-getxattr %v for atribute %v", fi.path, req.Name)
 
 	// debugLog("exec file getxattr: %v: %v", fi.path, req.Name)
-	xattrs, err := getXattr(fi.m.fs, req.Name, fi.path, req.Size)
+
+	// Do not worry about req.Size
+	// fuse will cut it to allowed size and report to the caller that buffer need to be larger
+	xattrs, err := getXattr(fi.m.fs, req.Name, fi.path)
 	if err != nil {
 		return err
 	}
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -113,6 +113,7 @@ func (fi *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.Open
 	if req.Flags.IsReadOnly() {
 		// we don't need to track read-only handles
 		// and no need to set handle `data`
+		resp.Flags |= fuse.OpenKeepCache
 		return fi.hd, nil
 	}
 
@@ -129,6 +130,7 @@ func (fi *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.Open
 		return nil, errorize(fi.path, ErrTooManyWriters)
 	}
 	fi.hd.writers++
+	resp.Flags |= fuse.OpenKeepCache
 	return fi.hd, nil
 }
 
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -34,6 +34,7 @@ type File struct {
 // Attr is called to get the stat(2) attributes of a file.
 func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 	defer logPanic("file: attr")
+	log.Debugf("fuse-file-attr: %v", fi.path)
 
 	info, err := fi.m.fs.Stat(fi.path)
 	if err != nil {
@@ -83,6 +84,7 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 func (fi *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenResponse) (fs.Handle, error) {
 	defer logPanic("file: open")
 	debugLog("fuse-open: %s", fi.path)
+	log.Debugf("fuse-file-open: %v with request %v", fi.path, req)
 
 	// Check if the file is actually available locally.
 	if fi.m.options.Offline {
@@ -135,11 +137,11 @@ func (fi *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.Open
 // file, the size change is noticed here before Open() is called.
 func (fi *File) Setattr(ctx context.Context, req *fuse.SetattrRequest, resp *fuse.SetattrResponse) error {
 	defer logPanic("file: setattr")
+	log.Debugf("fuse-file-setattr: request %v", req)
 
 	// This is called when any attribute of the file changes,
 	// most importantly the file size. For example it is called when truncating
 	// the file to zero bytes with a size change of `0`.
-	debugLog("exec file setattr")
 	switch {
 	case req.Valid&fuse.SetattrSize != 0:
 		if err := fi.hd.truncate(req.Size); err != nil {
@@ -158,8 +160,8 @@ func (fi *File) Setattr(ctx context.Context, req *fuse.SetattrRequest, resp *fus
 // Currently, fsync is completely ignored.
 func (fi *File) Fsync(ctx context.Context, req *fuse.FsyncRequest) error {
 	defer logPanic("file: fsync")
+	log.Debugf("fuse-file-fsync: %v", fi.path)
 
-	debugLog("exec file fsync")
 	return nil
 }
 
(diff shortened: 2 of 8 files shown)


catfs/mio/overlay/overlay.go
```diff
@@ -228,6 +228,24 @@ func (l *Layer) Write(buf []byte) (int, error) {
 	return len(buf), nil
 }
 
+// Writes data from `buf` at offset `off` counted from the start (0 offset).
+// Mimics `WriteAt` from `io` package https://golang.org/pkg/io/#WriterAt
+func (l *Layer) WriteAt(buf []byte, off int64) (n int, err error) {
+	// Copy the buffer, since we cannot rely on it being valid forever.
+	modBuf := make([]byte, len(buf))
```
sahib, line 235: Potential improvement here: use a [sync.Pool](https://golang.org/pkg/sync/#Pool) to save a few allocations when writing many buffers. This also applies to the regular ``Write`` method and is probably more relevant there. Care must be taken that the buffer from ``pool.Get()`` has a suitable length.

catfs/mio/overlay/overlay.go
```diff
@@ -342,6 +360,11 @@ func (l *Layer) Seek(offset int64, whence int) (int64, error) {
 		l.limit = newPos
 	}
 
+	if l.pos == newPos {
+		// very likely it sequent read/write request
+		// no need to seek since we already pointing to the right position
+		return l.pos, nil
+	}
```
sahib, line 367: :+1:

fuse/handle.go
```diff
@@ -113,32 +65,47 @@ const maxInt = int(^uint(0) >> 1)
 
 // Write is called to write a block of data at a certain offset.
 func (hd *Handle) Write(ctx context.Context, req *fuse.WriteRequest, resp *fuse.WriteResponse) error {
+	start := time.Now()
+	// Note: even when underlying process makes consequent writes,
+	// the kernel might send blocks out of order!!!
```
sahib: Can you elaborate a bit on this? Especially what the consequences for us are.

fuse/handle.go
```diff
@@ -113,32 +65,47 @@ const maxInt = int(^uint(0) >> 1)
...
 	hd.wasModified = true
-	resp.Size = n
 	return nil
 }
 
+// Writes data from `buf` at offset `off` counted from the start (0 offset).
+// Mimics `WriteAt` from `io` package https://golang.org/pkg/io/#WriterAt
+// Main idea is not bother with Seek pointer, since underlying `overlay` works
+// with intervals in memory and we do not need to `Seek` the backend which is very time expensive.
+func (hd *Handle) WriteAt(buf []byte, off int64) (n int, err error) {
```
sahib: I would recommend to lowercase the ``writeAt`` method here. I was confused at first because I thought that fuse expects this interface and would call it directly. I used lower case method names to also indicate that they don't do locking.

fuse/mount.go
```diff
@@ -67,6 +67,10 @@ func NewMount(cfs *catfs.FS, mountpoint string, notifier Notifier, opts MountOpt
 		fuse.FSName("brigfs"),
 		fuse.Subtype("brig"),
 		fuse.AllowNonEmptyMount(),
+		// enabling MaxReadahead double or even triple Read throughput 12MB/s -> 25 or 33 MB/s
+		fuse.MaxReadahead(128 * 1024), // kernel uses at max 128kB = 131072B
+		// enabling WritebackCache doubles write speed to buffer 12MB/s -> 24MB/s
+		fuse.WritebackCache(), // writes will happen in mach large blocks 128kB instead of 8kB
```
sahib, line 73: Nice nice nice.

fuse/util.go
```diff
@@ -46,7 +48,22 @@ func listXattr(size uint32) []byte {
...
+func isKnownAttribute(name string) bool {
+	reqSize := uint32(1024)
+	knownAttrs := bytes.Split(listXattr(reqSize), []byte{0})
+	for _, attr := range knownAttrs {
+		if string(attr) == name {
+			return true
+		}
+	}
+	return false
+}
```
sahib: I think this one could be faster. Calling ``getXattr`` often will always have the cost of calling ``listXattr`` (allocation heavy) and linearly checking for the attribute. Instead we should just use a map since we know and control the possible xattrs:

```go
var  xattrMap = map[string]bool{
    "user.brig.hash": true,
    ....
}

// use xattrMap in listXattr (iterate over keys) and getXattr (checking if a key exists)
```

Alternatively we could even think to do something like that:

```go

type xattrGetter func(cfs *catfs.FS, info catfs.StatInfo) ([]byte, error)

var xattrMap = map[string]xattrGetter{
    "user.brig.hash": func(cfs *catfs.FS, info catfs.StatInfo) ([]byte, error) {
        return []byte(info.TreeHash.B58String())
    },
    // ....
}


// Then the implementation of getXattr would be trivial:
func getXattr(cfs *catfs.FS, name, path string, size uint32) ([]byte, error) {
    // todo: not sure if it comes with a null byte or not.
    handler, ok := xattrMap[name]
    if !ok {
        return nil, fuse.ErrNoXattr
    }

    info, err := cfs.Stat(path)
    if err != nil {
        return nil, errorize("getxattr", err)
    }

    return handler(cfs, info)
}

// listxattr would be simpler too.
```

Advantage of this approach would be: lookup time of a map is constant and involves validating it in one go. Also the xattr names would be in one central place.

fuse/file.go
```diff
@@ -55,11 +56,7 @@ func (fi *File) Attr(ctx context.Context, attr *fuse.Attr) error {
 			attr.Mode = os.ModeSymlink | filePerm
 		}
 	}
-	if fi.hd != nil && fi.hd.writers > 0 {
-		attr.Size = uint64(len(fi.hd.data))
-	} else {
-		attr.Size = info.Size
-	}
+	attr.Size = info.Size
```
sahib, line 59: I'm not entirely sure, but ``info.Size`` might not be correct. When we truncate or extend a file then ``catfs`` still remembers the old size (it gets updated only after Stage(), i.e. on Flush()). We might need to use the limits of the overlay to determine the size at that point. At least that's what I remember, in any case we should check the file size on this test:

```bash
# file should have a size of 1MB.
$ dd if=/dev/zero of=/fuse/file.txt bs=1M count=1
# file size should increase with each write happening.
# (probably easier to test in a unit test where you can
#  control when the file is closed). 
$ dd if=/dev/zero of=/fuse/file.txt bs=1M count=20

```

sahib: This might also explains why your test produced that much metadata: Every fuse write called ``catfs.Handle.Write()`` and it stages the file to update the file size. That's actually something very dumb to do, and we should retrieve the current files ize from the handle and not from ``catfs`` (where it will be changed on flush).

sahib (changes requested):
> Enabling fuse optimizations at mount: Cache and larger buffer sizes. This gives quite large speed improvements, I even had to think about disabling debug outputs, since it takes comparable time to disk io.

Yep, logging is always surprisingly expensive.

> Main slow down for using overlay was useless Seek at write: we write to memory overlay but aks the backend to Seek for no reason. This problem solved with WriteAt method.

I think the Seek was pointless indeed. Probably a leftover from pre-overlay times. I think it could be also factored out of the normal ``handle.Write()`` method. On the other hand we don't use that otherwise, so might be removed completely too. Also, the old write did some metadata staging of the file size, which was very CPU intensive, so it was not only the Seek.

> @sahib please pay attention how the overlay limits are handled. I am not sure that I fully grasped what for they are used. The structure field had no comment :(

It's for the case that a file gets extended or truncated. Imagine a file that has 4KB, but then gets overwritten by a 1MB file. The file size in catfs would still be 4KB before Flush() was called, but fuse layer/overlay already knows that things got bigger. Same goes for the opposite of truncating a file (you can technically also make a file bigger by truncating, but that's a confusing term). The limit controls how big the file currently is. I left a comment about the file size (which was wrong before too), which we definitely need to check.

----

Otherwise very good. If overlay now learns to swap out excess intervals to disk overs a certain memory limit, then this would be a great improvement and would make the filesystem ready for day-to-day use.

fuse/handle.go
evgmik: > Can you elaborate a bit on this? Especially what the consequences for us are.

I will add the comment but the full description is the following:
let's say at the OS level you do `(cat or dd) largeFile > /brig-fuse-moung/file`. You would naively expect `Write` request to be in `fifo` order with steady growing offsets (I called it consequent), and they mostly are.  What I saw that occasionally OS layer asks to write a block with higher offset and then another with a lower offset.

I noticed this when I was tracking offset position: with `fifo` order I would not have to do additional `Seek` if position matches offset.

It does not concern our current `Write` much. But one should not assume that `Write` could be streamed with automatically positioned cursor.
Otherwise, it would be super easy to merge `overlays` to one big `modification` which might help with speed. 

But it also makes, aforementioned swapping to disk hard 'since overlay might have a hole'. So every block would be of a different size.

evgmik:
> Yep, logging is always surprisingly expensive.
Well, everyone who worked with micro controllers now that a `printf` will puts everything to the dead snail speed :)
I was surprised to see, that it still true.

I looked at other loggers [Zerolog](https://github.com/rs/zerolog) and [Zap](https://github.com/uber-go/zap). It seems that they get their speed from extremely rigid formatting. I.e. you cannot day `Printf("%v", value)`, you have to spell out the type.
Which makes it painful for sprinkling debugging info.

> > @sahib please pay attention how the overlay limits are handled. I am not sure that I fully grasped what for they are used. The structure field had no comment :(
> 
> It's for the case that a file gets extended or truncated. Imagine a file that has 4KB, but then gets overwritten by a 1MB file. The file size in catfs would still be 4KB before Flush() was called, but fuse layer/overlay already knows that things got bigger. Same goes for the opposite of truncating a file (you can technically also make a file bigger by truncating, but that's a confusing term). The limit controls how big the file currently is. I left a comment about the file size (which was wrong before too), which we definitely need to check.
> 
The confusing part is  
https://github.com/sahib/brig/blob/c29348532553a533edee7e249574b48b8cb51f23/catfs/mio/overlay/overlay.go#L194
there are `limit` and `fileSize`. From your description it sounds that they should be the same.

evgmik:
> If overlay now learns to swap out excess intervals to disk overs a certain memory limit, then this would be a great improvement and would make the filesystem ready for day-to-day use.

I know that it is one of your needed features list. But after thinking about it, I think it should be pushed to future or even abandoned.

I am thinking about use cases:
1. `brig` is a syncer
 - User would copy a file (maybe a very large one) in a single command. Yes, we ask the size of the file to be allocated in memory, (and about the same amount in OS cache).
   This is a *transient* memory demand, once we flushed the `overlay` the memory is free. Moving to swap would cost us speed.
2. User edits small file directly in `brig-fuse`. Small files not a big deal.
3. `brig-fuse` is the real file system.
  - I.e. someone might run a database with random access read/writes. I see no efficient way to do it, which would not do the high `io` to the harddrive for swapping.
  - The only easy way to do it, is to make backend blocked,  instead of the current mode: one file one hash. I guess there is a reason why real FS split files on multiple inodes.

I would argue that case 3 is not `brig` use case. Doing efficient FS is hard work (I still remember sqlitedb speed drop on transition from ext2 to ext3 due to some journaling). It is potentially doable: internal streamed data is prepared in chunks (archived and encrypted).
But then we are reinventing `ipfs` file storage. 


Case 1, would greatly benefit, if there were a flag from OS saying that what is coming is the single "stage" but I see no such thing in the manuals.

catfs/mio/overlay/overlay.go
evgmik: Not sure that I follow here. `modBuf` is somewhat permanent (sits in the `overlay` until flashed). Pool implies, that the data would be soon deleted. Besides manual says "Any item stored in the Pool may be removed automatically at any time without notification.". This sounds very scary.

I guess I do not know internals of `gc` to really see the difference.

fuse/handle.go
evgmik: addressed in d3aad7af

fuse/handle.go
evgmik: done in 03e48e64

sahib:
> there are limit and fileSize. From your description it sounds that they should be the same.

No, as described above they're not. ``limit`` should be the actual size of the file, ``fileSize`` is the size of the original file before it was opened in fuse. Disclaimer: I didn't re-read the code, that's just my faulty memory, so reality might be different or buggy.

But in your PR you removed the part were on every write the file was re-staged with the up-to-date size. While this removal is good, the file size reported by fuse needs to stay up-to-date on every Write. You gonna need to expose it from overlay to handle to fuse's getattr.

> I know that it is one of your needed features list. But after thinking about it, I think it should be pushed to future or even abandoned.

Your point 1 is already the reason why we need this. Without it we can only copy files up to the size of our memory. The limit when we start swapping to disk can be set high enough that it does not happen for normal usage.

I don't expect anybody to run a database on a fuse, but people might still edit large files for whatever reason. This may not be a typical use case and we don't need to be fast here, but we still need to support it to a certain extent.

**EDIT:** Clarfication: We can of course postpone it and it shouldn't be part of that PR anyways.

catfs/mio/overlay/overlay.go
sahib: That documentation is a little misleading. Nothing gets garbage collected while you still use it (i.e. you hold a reference to it). ``modBuf`` can be ``Put()`` back to the pool after flush and on L235 above you can do a ``pool.Get()`` to fetch a buffer that was either used before or is newly allocated. Go is free to do whatever it likes in the time between Put and Get.

evgmik: I see now. Can we move it to separate issue?
 From quick reading, `pool` is not suited for different size objects. So if `len(buf)` different from result of a `Get`, we would have to put it back. But I see no way, to iterate over `pool` to see if there is a suitable candidate. Besides, it would require the full rethinking of how `overlay` is working.

evgmik:
> > there are limit and fileSize. From your description it sounds that they should be the same.
> 
> No, as described above they're not. `limit` should be the actual size of the file, `fileSize` is the size of the original file before it was opened in fuse. Disclaimer: I didn't re-read the code, that's just my faulty memory, so reality might be different or buggy.
> 
> But in your PR you removed the part were on every write the file was re-staged with the up-to-date size. While this removal is good, the file size reported by fuse needs to stay up-to-date on every Write. You gonna need to expose it from overlay to handle to fuse's getattr.
> 
> > I know that it is one of your needed features list. But after thinking about it, I think it should be pushed to future or even abandoned.
> 
> Your point 1 is already the reason why we need this. Without it we can only copy files up to the size of our memory. The limit when we start swapping to disk can be set high enough that it does not happen for normal usage.
> 
> I don't expect anybody to run a database on a fuse, but people might still edit large files for whatever reason. This may not be a typical use case and we don't need to be fast here, but we still need to support it to a certain extent.
> 
> **EDIT:** Clarfication: We can of course postpone it and it shouldn't be part of that PR anyways.

Let's postpone this particular point. The OS keeps track of the file size internally, since it does the writes, it knows better what the file size is. How I check, I `dd`ed 1GB file to `brig-fuse`, which takes a while, and I was doing `ls fuse/file` during `dd`ing. I saw that file size was growing in sync with write requests.

evgmik (commented):
OK. Now I see your point. Big files, on small memory machines. Yep, this should be addressed.
But I suggest to move it to different issue.

My particular PR, does not change current memory demand, so it just maintains status quo.

fuse/util.go
evgmik: I like it. Mostly for unified track of available attributes. Speed wise, it is probably will gain as very little with only 3 attributes. Preparing patch now.

evgmik:
What is wrong with travis? All test are passed on my machine.

fuse/util.go
sahib: The actual speed-up would only be noticeable when doing a lot of xattr fetching. But this would both improve implementation and speed, so a win-win.

sahib:
> How I check, I dded 1GB file to brig-fuse, which takes a while, and I was doing ls fuse/file during dding. I saw that file size was growing in sync with write requests.

That's unexpected. My understanding is that it still makes a call to ``getattr()`` to get the size. Will double check this tomorrow.

>  [...] But I suggest to move it to different issue.

Sure, that needs another PR. I would keep #49 open since it's a long running issue with some more thing to do.
If you want I can continue improving ``overlay`` - since I brought that beast into life anyways.

> What is wrong with travis? All test are passed on my machine.

Good question. ``TestClientFetchPatch`` broke for some reason. Maybe a flaky test. I re-run now and it's green now.
Can you see the logs [here](https://travis-ci.org/github/sahib/brig/builds/755866780)? Can you also re-trigger the build with the "build now" button?

sahib:
jkjk

sahib:
Oops, accidentally hit the close button.

catfs/mio/overlay/overlay.go
sahib: You would just get an item from the pool and check if it's reasonably sized (if larger, cut it), if not allocate a new one the normal way and use that. But sure, we can do it once overlay is touched in general.

evgmik:
> Sure, that needs another PR. I would keep #49 open since it's a long running issue with some more thing to do.
> If you want I can continue improving `overlay` - since I brought that beast into life anyways.

I would greatly appreciate it, I think I understand what it *should* do, but internal machinery does not fit yet to my brain.

> > What is wrong with travis? All test are passed on my machine.
> 
> Good question. `TestClientFetchPatch` broke for some reason. Maybe a flaky test. I re-run now and it's green now.
> Can you see the logs [here](https://travis-ci.org/github/sahib/brig/builds/755866780)? Can you also re-trigger the build with the "build now" button?

I see logs, not sure that I have `rebuild` button. It passed the checks, so it is probably disabled.

fuse/util.go
evgmik: done with c940779f, I also fixed old unnoticed bug with ce915cef and added a test case in 05b57370

I think we are ready to merge

sahib (approved):
Looks good to me.

Issue #20, opened 2018-12-09
Provide better autocompletion


sahib:
Many of the current commands (like ``brig stg``) have subpar quality in auto completion.
``brig stg`` for example only completes files in the current local directory. This is could be a lot smoother, but needs some bash/zsh trickery.

sahib:
This would be also a good idea to be worked upon right now. [This](https://iridakos.com/tutorials/2018/03/01/bash-programmable-completion-tutorial.html) looks like a good tutorial.

PR #90, opened 2021-01-24
Feature/hints

sahib:
This is not finished yet, but I wanted to push the thing for a general overview. You don't need to review it yet, just in case you're curious. Things are pretty buggy right now, but that's just a bit of cleanup I need to do. The gist of this PR is:

* Add a new ``brig hints`` command (+ subcommands) that allow configuring compression and/or encryption algorithm for files and directories. Hints are only used on the next stage and they are called "hints" because we can ignore them in future versions.
* Add ``IsRaw`` attribute to file nodes to indicate that this file could be streamed from IPFS directly (also used during streaming of file to figure out if we need to use any of our decoding). Also an interesting for the gateway: Those files can be hyperlinked directly by appending the BackendHash to a ``ipfs`` gateway.
* Teach ``mio.Stream`` to read hints and the isRaw attribute. This required some changes to the encryption header, which is a backwards-incompatible change. Files encoded with older versions will probably spit out errors (didn't bother to check).
* Add ``brig stage --recode`` flag to recode files when you changed hints. This will work for directories too, but does not yet.


catfs/mio/compress/algorithm.go
```diff
@@ -9,49 +9,74 @@ import (
...
 var (
 	// ErrBadAlgo is returned on a unsupported/unknown algorithm.
-	ErrBadAlgo = errors.New("Invalid algorithm type")
+	ErrBadAlgo = errors.New("invalid algorithm type")
 )
 
+const (
+	// AlgoUnknown represents an unknown algorithm.
+	// When trying to use it an error will occur.
+	AlgoUnknown = AlgorithmType(iota)
```
evgmik, line 18: `AlgoUnknown` is not the same as `AlgoNone`. Shall we bring back `AlgoNone` as well?

catfs/mio/compress/common.go
```diff
@@ -120,12 +85,16 @@ func makeHeader(algo AlgorithmType, version byte) []byte {
 	return append([]byte("elchwald"), suffix...)
```
evgmik: `MagicNumber` ?

catfs/mio/encrypt/format.go
```diff
@@ -32,17 +32,35 @@ import (
...
+	// This should be fast on modern CPUs.
+	FlagEncryptAES256GCM = Flags(1) << iota
+
+	// FlagEncryptChaCha20 incidate that the stream was encrypted with ChaCha20.
+	// This can be a good choice if your CPU does not support the AES-NI instruction set.
+	FlagEncryptChaCha20
+
+	// FlagCompressedInside indicates that the encrypted data was also compressed.
+	// This can be used to decide at runtime what streaming is needed.
+	FlagCompressedInside
```
evgmik, line 71: Shall we leave room for possible new encryption algorithm, and push bit 1 more to the left for `FlagCompressedInside`, after all we have 32 bits to spare

evgmik (approved):

catfs/mio/compress/algorithm.go
sahib: No, because one wants to use ``AlgoNone``, then the ``compress`` package should not be used. For performance reasons I wanted to avoid stacking readers. Old ``AlgoNone`` e.g. still wrote all the index structure and chunked the data, but just did not compress it. The performance impact was mild but still there.

catfs/mio/compress/common.go
sahib: Oops, merci.

catfs/mio/encrypt/format.go
sahib: We could also add new encrypt flags after ``FlagCompressedInside``, but I agree it might be tidier to move it back a bit.

sahib:
FYI: Did some fixes yesterday, but then tried to run the tests which filled up my memory and I had to reboot. Last update broke something though, so I have to fix that first. :man_facepalming: 

EDIT: Okay, after a happy round of GRUB re-installing I'm back.

catfs/fs.go
```diff
@@ -1006,7 +1006,7 @@ func (fs *FS) Stage(path string, r io.Reader) error {
 		hint = fs.hintFetcher.Lookup(path)
 	}
 
-	stream, err := mio.NewInStream(sizeReader, path, key, hint)
+	stream, isRaw, err := mio.NewInStream(sizeReader, path, key, hint)
```
evgmik, line 1023: Would it be better to have the `isRaw` field in the stream structure? Some of our stream already have extra info and it is convenient to keep it in one place.

evgmik (approved):

catfs/fs.go
sahib: I made it a return parameter because it's something that needs to be stored persistently outside of the stream (in this case as part of the metadata). Making it a part of stream would suggest otherwise. Another downside is that stream often is passed through as ``io.Reader`` or ``io.ReadSeeker``. This would strip access to the ``isRaw`` attribute from the stream.

evgmik (approved):

sahib:
Still work in progress, but you might be delighted by da3b3c04. Now your system should use ``ChaCha20`` by default, while my Intel laptop stays on AES. Please check if that heuristic works. Otherwise some more tests and documentation missing, otherwise the PR should be good to go. Probably another day of work.

docs/features.rst
```diff
@@ -37,14 +45,16 @@ Truly Distributed
...
 * Simple user identification and discovery.
-* You do not store data you don't want to.
+* You do not store data you don't want to store. Pin what you need,
+  fetch the rest from the network on request.
 
 FUSE filesystem
 ---------------
 
-* FUSE filesystem that makes all files seem like a normal directory.
+* FUSE filesystem that acts your data is in a local directory.
```
evgmik: > FUSE filesystem that acts your data is in a local directory.

This sounds strange in English. How about:

"FUSE filesystem mirrors your data to a directory."

docs/tutorial/coreutils.rst
```diff
@@ -80,3 +80,82 @@ the available commands:
...
+
+Available compression algorithms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. todo:: Provide relative throughput numbers compare to none.
+
++----------------+----------------------------------------------------------------------------+
+| NAME           |   DESCRIPTION                                                              |
++================+============================================================================+
+| ``snappy``     | High throughput, relative low compression rate.                            |
```
evgmik: compression rate -> compression ratio

docs/tutorial/coreutils.rst
```diff
@@ -80,3 +80,82 @@ the available commands:
...
++----------------+----------------------------------------------------------------------------+
+| ``lz4``        | High throughput, slightly higher compression rate than snappy.             |
```
evgmik: rate -> ratio

docs/tutorial/coreutils.rst
```diff
@@ -80,3 +80,82 @@ the available commands:
...
++----------------+----------------------------------------------------------------------------+
+| ``guess``      | Choose best algorithm based on file ending, size and mime type.            |
```
evgmik: > Choose best algorithm based on file ending, size and mime type.

Chooses the best algorithm based on a file ending, size and mime type.

evgmik (approved):

docs/tutorial/coreutils.rst
```diff
@@ -146,16 +146,16 @@ Available encryption algorithms
...
 | NAME           |   DESCRIPTION                                                              |
 +================+============================================================================+
-| ``snappy``     | High throughput, relative low compression rate.                            |
+| ``snappy``     | High throughput, relative low compression ratio.                           |
 +----------------+----------------------------------------------------------------------------+
-| ``lz4``        | High throughput, slightly higher compression rate than snappy.             |
+| ``lz4``        | High throughput, slightly higher compression ratio than snappy.            |
 +----------------+----------------------------------------------------------------------------+
-| ``guess``      | Choose best algorithm based on file ending, size and mime type.            |
+| ``guess``      | Chooses best algorithm based on file ending, size and mime type.           |
```
evgmik: I thought about it a bit more. The "*the* best" implies the absolute best. While here we are actually guessing based on prior knowledge.

How about:
"Picks a compression algorithm based on file extension, size, and mime type."

By the way do we really use `size`? It is used as a threshold to do compression or not, but I think it is uniform for all algorithm. I.e. it would be enough to say somewhere above "Files smaller than ... are not compressed"

sahib: > How about:
> "Picks a compression algorithm based on file extension, size, and mime type."

I replaced "best" with "suitable".

> By the way do we really use size? It is used as a threshold to do compression or not, but I think it is uniform for all algorithm. I.e. it would be enough to say somewhere above "Files smaller than ... are not compressed"

Sure, but we might use the size at some point. Some compression formats have higher overhead for each block (zstd vs zlib e.g.). There it makes sense to choose another algorithm until a certain size. For the user it's not that important what attributes we actually use for the heuristic and how exactly the guessing works. From my experience it's good to not go into detail too much in documentation. That also saves us from updating (or forgetting to do so) whenever we change the heuristic.

evgmik: I like your reasoning. Let's go with it.

sahib:
@evgmik: I'm more or less done with that PR. I might add a few tests, but no functional changes planned. Would be good if you could have a play with the feature and give feedback, if any. Maybe I oversaw something.

evgmik:
> @evgmik: I'm more or less done with that PR. I might add a few tests, but no functional changes planned. Would be good if you could have a play with the feature and give feedback, if any. Maybe I oversaw something.

I played with it a bit, here my thoughts:

- Debug info in client: "INFO[0000] using chacha20 as default because CPU does not support AES-NI" is very annoying. It shows even if I do simple `ls` on a directory and breaks auto completions.
- Also it seems that we broke auto completions, I recall that `brig ls /REA<tab>` use to autocomplete filenames and directories. But this is not working even in `develop` branch

Bigger items
 - I think hints display should be merged with `info` and `ls`. I.e. if I do `brig info file` it should show me hints, it seems to be quite natural. Same for `brig ls file`, we should add `compression` and `encryption` columns
 - Currently `brig hints ls ` is very counter intuitive. It shows only root directory `/` hints and additionally only the directories with not default hints.
 - `brig hint ls` does not show files!
 - I see no way to see hints for a file
 - `brig stage --recode` is very unusual command, I would not find it if I did not know about it. I think it should have its own `brig recode`.
 - hints are somewhat weird, they show what you want and not what it is. Not sure what to do with it. But I would say change of a hint should trigger recursive `recode`.

Not tested yet, but we need to think how hints transfer at synchronization. I.e. which settings win.

sahib:
Thanks for the feedback.

> Debug info in client: "INFO[0000] using chacha20 as default because CPU does not support AES-NI" is very annoying. It shows even if I do simple ls on a directory and breaks auto completions.

Oops, didn't realize it's also happening in client. That probably also bad for the auto completion (there might be more things that are wrong). Improving the auto-completion is a also a thing that definitely needs improvements. It sucks a lot currently, even if it's working. Would be a huge step regarding usability. We already have a ticket [here](https://github.com/sahib/brig/issues/20).

> I think hints display should be merged with info and ls. I.e. if I do brig info file it should show me hints, it seems to be quite natural. Same for brig ls file, we should add compression and encryption columns

Agreed, I will add it to the info and ls command. I wondering if it should be shown by default in ls, or by adding an option...

> Currently brig hints ls is very counter intuitive. It shows only root directory / hints and additionally only the directories with not default hints.
> brig hint ls does not show files!
> I see no way to see hints for a file

That's on purpose. A user won't probably give hints for each file and we're only showing the one that he explicitly set. I agree though that we should show what hints are applied for specific files. But that should be fixed by adding hints to ``brig info`` and ``brig ls``.

> brig stage --recode is very unusual command, I would not find it if I did not know about it. I think it should have its own brig recode.

Agreed, don't really know why I put it there. I think I would add a ``brig hints recode`` command and a ``--recode`` option to ``brig hints set`` that change the encoding. Wouldn't make it a top-level command since it's useful in conjunction with hints and we're having already too many top-level commands.

> hints are somewhat weird, they show what you want and not what it is. Not sure what to do with it. But I would say change of a hint should trigger recursive recode.

I definitely would not make automatic recoding a default. As written above, we can add a ``--recode`` option, but the nice part is that user can set the hint and that it gradually gets enforced while working without wasting any additional resources. If you want to see immediate results, you can do that. We can later on add commands to ask brig what encoding a file currently has, but I don't consider that important for now.

> Not tested yet, but we need to think how hints transfer at synchronization. I.e. which settings win.

Nobody wins. Files stay as they are, except remote side modifies them, then their settings get applied. Hints do not get synchronized on purpose (also a point why they're not in the nodes), since it's a setting specific to the machine where the brig repository is located.


-----

On another note, I think I will tackle the benchmarking command after this PR, since I consider it as a building block for the next changes (new compression algorithms, overlay disk swapping).

evgmik (approved):

evgmik (approved):

evgmik:
I love new way with `brig ls`, It all makes sense. 

There is a bug at hints set with --recode switch. The hints are not
set in this case.

```
$ brig hi set  -r /README.md --compression none --encryption none
$ brig ls /README.md                         (1 sec)
SIZE   BKEND  MODTIME                  PATH        PIN  CACHED  HINT
645 B  720 B  2021-01-31 12:06:08 EST  /README.md   ✔     ✔     enc:chacha20-zip:guess
```

when executed without `-r` switch, everything as it should
```
$ brig hi set /README.md --compression none --encryption none
$ brig ls /README.md
SIZE   BKEND  MODTIME                  PATH        PIN  CACHED  HINT
645 B  720 B  2021-01-31 12:06:08 EST  /README.md   ✔     ✔     enc:none-zip:none
```

sahib:
Interesting. That seems to be a bug in the ``cli`` library. When I add any string at the end of the command or if I move the ``--recode`` past the path argument, then it works. :open_mouth: Otherwise, the arguments for compression and encryption flags are not recognized because parsing seems to stop after ``--recode`` (``ctx.Args()`` contains them). Because of that the value is not changing. Really strange.

I added a "work around" for that in fd3bd4dc, so that we at least notice when the command did not work at all. Someone need to dig deeper on how that happens.

evgmik:
> Interesting. That seems to be a bug in the `cli` library. When I add any string at the end of the command or if I move the `--recode` past the path argument, then it works. open_mouth Otherwise, the arguments for compression and encryption flags are not recognized because parsing seems to stop after `--recode` (`ctx.Args()` contains them). Because of that the value is not changing. Really strange.
> 
> I added a "work around" for that in [fd3bd4d](https://github.com/sahib/brig/commit/fd3bd4dcd4d7f1051403ebce4bc96a280df9c35b), so that we at least notice when the command did not work at all. Someone need to dig deeper on how that happens.

It seems to fix the problem. :+1: 

One more thing, is to have a list of possible encryption and zip algorithms via the command line. Right now, one has to read the code to have an idea about it.

sahib:
> One more thing, is to have a list of possible encryption and zip algorithms via the command line. Right now, one has to read the code to have an idea about it.

Well, they're also on the HTML docs, but I added them now too in the help text as of af2ef689.

evgmik:
> Well, they're also on the HTML docs, but I added them now too in the help text as of [af2ef68](https://github.com/sahib/brig/commit/af2ef689cada02cb68a0c20d122eb4f9237dc773).

Excellent work!

`fuse` layer seems to behave correctly too! I can set and read hints just fine.

sahib:
> fuse layer seems to behave correctly too! I can set and read hints just fine.

Good. I just realized though that setting hints was only possible for non-directories. Implemented that now as of 22fee5bd.
That feature can be extended to control pinning over xattrs. But I leave that to another changeset, since it's not related to hints.

Issue #83, opened 2021-01-12
Implement live benchmark command


evgmik:
Implement "live" benchmarks with something like `brig debug iobench`.
The idea was born in discussion of #82 

The idea is to test througput for
- for things like `brig cat` and `brig stage`
- fuse layer
- isolated benchmarks like encryption or compression

We could also output various system information that influences the benchmark results (CPU model, IPFS version etc.). All throughput numbers can also be expressed additionally as relative to the baseline (e.g. just copying bytes from ten-source to ten-sink).

sahib:
Closing since it was covered extensively.

PR #93, opened 2021-02-07
Feature/benchmark

sahib:
(NOTE: Still a draft, some cleanup is still required)

I had finally some time this weekend to implement a ``iobench`` command as it was outlined it in #83. This is now available as ``brig debug iobench`` and offers pretty much everything we'd want to measure regarding our streaming system, including fuse, server and ipfs benchmarks.

I factored it out into its own utility to make it easier accessible and scriptable. The output can be also JSON, which can then be stored and compared to older/newer versions. Here is an example on how this looks for benchmarking the ``mio`` package.
The progressbar that is shown indicates the relative throughput to the "baseline" (which is ``enc:none-zip:none`` in that case):

```
$ brig debug iobench --bench memcpy --bench mio-read --bench mio-write --size 256MB

Time:          2021-02-07T23:01:44+01:00
CPU Name:      Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
Logical Cores: 4
Has AESNI:     yes


ten:memcpy
==========

memcpy:ten_enc:none-zip:none                       [============================================================] 8445.58 MB/s (  100.00%)   30ms for 256.00 MB

ten:mio-read
============

mio-read:ten_enc:none-zip:none                     [============================================================] 6321.69 MB/s (  100.00%)   40ms for 256.00 MB
mio-read:ten_enc:none-zip:lz4                      [======______________________________________________________] 617.38  MB/s (    9.77%)  415ms for 256.00 MB
mio-read:ten_enc:none-zip:snappy                   [====================________________________________________] 2101.92 MB/s (   33.25%)  122ms for 256.00 MB
mio-read:ten_enc:none-zip:guess                    [====================________________________________________] 2084.60 MB/s (   32.98%)  123ms for 256.00 MB
mio-read:ten_enc:aes256gcm-zip:none                [================____________________________________________] 1637.54 MB/s (   25.90%)  156ms for 256.00 MB
mio-read:ten_enc:aes256gcm-zip:lz4                 [========____________________________________________________] 741.15  MB/s (   11.72%)  345ms for 256.00 MB
mio-read:ten_enc:aes256gcm-zip:snappy              [===================_________________________________________] 1997.19 MB/s (   31.59%)  128ms for 256.00 MB
mio-read:ten_enc:aes256gcm-zip:guess               [===================_________________________________________] 1927.34 MB/s (   30.49%)  133ms for 256.00 MB
mio-read:ten_enc:chacha20-zip:none                 [=============_______________________________________________] 1346.39 MB/s (   21.30%)  190ms for 256.00 MB
mio-read:ten_enc:chacha20-zip:lz4                  [=======_____________________________________________________] 691.78  MB/s (   10.94%)  370ms for 256.00 MB
mio-read:ten_enc:chacha20-zip:snappy               [================____________________________________________] 1595.05 MB/s (   25.23%)  160ms for 256.00 MB
mio-read:ten_enc:chacha20-zip:guess                [=================___________________________________________] 1767.81 MB/s (   27.96%)  145ms for 256.00 MB

ten:mio-write
=============

mio-write:ten_enc:none-zip:none                    [============================================================] 5236.99 MB/s (  100.00%)   49ms for 256.00 MB
mio-write:ten_enc:none-zip:lz4                     [===_________________________________________________________] 245.31  MB/s (    4.68%) 1.044s for 256.00 MB
mio-write:ten_enc:none-zip:snappy                  [===================_________________________________________] 1647.81 MB/s (   31.46%)  155ms for 256.00 MB
mio-write:ten_enc:none-zip:guess                   [===================_________________________________________] 1649.83 MB/s (   31.50%)  155ms for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:none               [=============_______________________________________________] 1116.71 MB/s (   21.32%)  229ms for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:lz4                [===_________________________________________________________] 223.93  MB/s (    4.28%) 1.143s for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:snappy             [===============_____________________________________________] 1248.39 MB/s (   23.84%)  205ms for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:guess              [==============______________________________________________] 1196.07 MB/s (   22.84%)  214ms for 256.00 MB
mio-write:ten_enc:chacha20-zip:none                [==========__________________________________________________] 811.42  MB/s (   15.49%)  315ms for 256.00 MB
mio-write:ten_enc:chacha20-zip:lz4                 [====________________________________________________________] 262.53  MB/s (    5.01%)  975ms for 256.00 MB
mio-write:ten_enc:chacha20-zip:snappy              [====================________________________________________] 1705.61 MB/s (   32.57%)  150ms for 256.00 MB
mio-write:ten_enc:chacha20-zip:guess               [================____________________________________________] 1314.64 MB/s (   25.10%)  195ms for 256.00 MB

```

The full list of available benchmarks is here:

```
$ brig debug iobench list
brig-read-ipfs:random
brig-read-ipfs:ten
brig-read-mem:random
brig-read-mem:ten
brig-write-ipfs:random
brig-write-ipfs:ten
brig-write-mem:random
brig-write-mem:ten
fuse-read-ipfs:random
fuse-read-ipfs:ten
fuse-read-mem:random
fuse-read-mem:ten
fuse-write-ipfs:random
fuse-write-ipfs:ten
fuse-write-mem:random
fuse-write-mem:ten
ipfs-read:random
ipfs-read:ten
ipfs-write:random
ipfs-write:ten
memcpy:random
memcpy:ten
mio-read:random
mio-read:ten
mio-write:random
mio-write:ten
```

I still need to verift the benchmarks actually measure what we want. The throughput in the example above is very high because we really only measure the write or read and immediately throw away the written or read block. You won't get those number from doing ``brig cat``, but they are easier to compare to each other (no memory allocation needed, less dependent on speed of RAM).

If you decide to run all of them you should tune (i.e. lower) the ``--size`` parameter. Since everything happens in ``/tmp``, you might experience an emergency reboot as I did earlier. :smirk: 

----

An interesting change could be that I implemented a new package ``fuse/fusetest`` that can be used to spin up an own process that serves a FUSE filesystem (either backed by memory or IPFS, but always catfs). Implemented this freshly, because the util from ``bazil/fuse`` could only be used with the testing package. Currently this is only used by the ``bench`` package, but adapting/extending it to the test should be rather easy. It turns the roles around: The tests could then do the file I/O themselves while the fuse mount is being served from a process started per test.


evgmik:
Wow. This is fantastic!!! Looking at yours and mine benchmark. I see a common theme: `lz4` is horrendous from the throughput point of view. I am a bit surprised by it, I thought that at least reading should be fast.

Here is another surprise from my benchmark (same parameters as yours):
```
mio-write:ten_enc:aes256gcm-zip:none               [==                                                          ] 49.94   MB/s (    1.70%) 5.126s for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:lz4                [====                                                        ] 179.78  MB/s (    6.10%) 1.424s for 256.00 MB
mio-write:ten_enc:aes256gcm-zip:snappy             [============                                                ] 573.99  MB/s (   19.48%)  446ms for 256.00 MB
```

How come that `mio-write:ten_enc:aes256gcm-zip:none` is the slowest?
(EDIT) I see, this is because uncompressed stream is longer and the actual streams order is
`mio-write:ten -> zip:none - >enc:aes256gcm`. I would suggest to make output of the benchmark somewhat reflecting the order.

Another enhancement, is to output the bar graph as the very last. It does not fit my regular `80 char` screens, as well as github window. But when it is the last, most of the time I would see enough.

sahib:
>  I would suggest to make output of the benchmark somewhat reflecting the order.

The hint is just encoded in the name. I could swap it around, but since it's an internal util I don't see much issue once we learned that compression influences the size of data (also took me a while the first time :smile:).

> Another enhancement, is to output the bar graph as the very last. It does not fit my regular 80 char screens, as well as github window. But when it is the last, most of the time I would see enough.

Changed it now to scale with the terminal width. Just a quick heuristic, but should work well enough.

----

I think there are some benchmark (especially the IPFS ones) that also produce rather surprising values. I think they are partly wrong due to setup costs of IPFS/fuse. Have to check, but otherwise it's starting to look good. At some point later in time we can write a bit of tooling to generate graphs from the results.

cmd/iobench.go
```diff
@@ -135,24 +136,37 @@ func drawHeading(heading string) {
...
+		w = 100
+	}
+
 	perc := float64(ref) / float64(took)
-	const cells = 60
 
-	fmt.Printf("%-50s [", name)
+	// take a guess on how big the bar may be.
+	var cells int
+	if barw := w - 40 - 40; barw < cells {
```
evgmik: This heuristics fails on my terminal which 101 symbols wide. I would still argue to put bar plot at the end,
since we have good control over throughput and algorithm fields.

evgmik (approved):

evgmik:
> I think there are some benchmark (especially the IPFS ones) that also produce rather surprising values. I think they are partly wrong due to setup costs of IPFS/fuse. Have to check, but otherwise it's starting to look good. At some point later in time we can write a bit of tooling to generate graphs from the results.

What triggered your attention, I did quick run on fuse and IPFS, they seems to behave as expected (HDD speed limited).

cmd/iobench.go
sahib: I removed the bar completely now since it did not produce so much value by itself. The heuristic you commented on was wrong by the way - it was fixed in the commit after.

sahib:
> What triggered your attention, I did quick run on fuse and IPFS, they seems to behave as expected (HDD speed limited).

I guess you had a running ipfs daemon? Without the benchmarks starts a new one and in the first few seconds the performance is really bad but normalizes soon after. I need to double check. Maybe makes sense to always start an ipfs instance (in /tmp) and disable online features in the config.

sahib:
Just for reference, those are the numbers I get for the two benchmarks that seem suspicious to me:

```bash
$ brig debug iobench -b brig-read-ipfs --size 100M


Time:          2021-02-09T23:14:53+01:00
CPU Name:      Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
Logical Cores: 4
Has AESNI:     yes


ten:brig-read-ipfs
==================

brig-read-ipfs:ten_enc:none-zip:none          216.36   MB/s       100.00MB/462ms   100.00%
brig-read-ipfs:ten_enc:none-zip:lz4           30.87    MB/s       100.00MB/3.24s    14.27%
brig-read-ipfs:ten_enc:none-zip:snappy        11.51    MB/s      100.00MB/8.689s     5.32%
brig-read-ipfs:ten_enc:aes256gcm-zip:none     226.23   MB/s       100.00MB/442ms   104.56%
brig-read-ipfs:ten_enc:aes256gcm-zip:lz4      522.24   MB/s       100.00MB/191ms   241.38%
brig-read-ipfs:ten_enc:aes256gcm-zip:snappy   953.93   MB/s       100.00MB/105ms   440.90%
brig-read-ipfs:ten_enc:chacha20-zip:none      230.62   MB/s       100.00MB/434ms   106.59%
brig-read-ipfs:ten_enc:chacha20-zip:lz4       525.26   MB/s       100.00MB/190ms   242.77%
brig-read-ipfs:ten_enc:chacha20-zip:snappy    775.17   MB/s       100.00MB/129ms   358.28%

ten:fuse-read-ipfs
==================

fuse-read-ipfs:ten_enc:none-zip:none          4360.01  MB/s        100.00MB/23ms   100.00%
fuse-read-ipfs:ten_enc:none-zip:lz4           4597.24  MB/s        100.00MB/22ms   105.44%
fuse-read-ipfs:ten_enc:none-zip:snappy        3946.10  MB/s        100.00MB/25ms    90.51%
fuse-read-ipfs:ten_enc:aes256gcm-zip:none     4136.77  MB/s        100.00MB/24ms    94.88%
fuse-read-ipfs:ten_enc:aes256gcm-zip:lz4      4380.45  MB/s        100.00MB/23ms   100.47%
fuse-read-ipfs:ten_enc:aes256gcm-zip:snappy   4103.91  MB/s        100.00MB/24ms    94.13%
fuse-read-ipfs:ten_enc:chacha20-zip:none      4197.01  MB/s        100.00MB/24ms    96.26%
fuse-read-ipfs:ten_enc:chacha20-zip:lz4       3376.74  MB/s        100.00MB/30ms    77.45%
fuse-read-ipfs:ten_enc:chacha20-zip:snappy    4546.68  MB/s        100.00MB/22ms   104.28%
```

* ``brig-read-ipfs`` does not make sense. There's no reason why ``enc:none-zip={snappy,lz4}`` should be this slow compared to the ones with encryption. Probably some bug, but too tired to check now.
* ``fuse-read-ipfs`` yields numbers higher than ``mio-read`` which is... surprising. I also suspect a bug here.

EDIT: ``ipfs`` doesn't seem to be the issue. The issue persists also when randomizing the benchmark order. I still implemented a way to start it in test mode to minimize network traffic and unrelated noise.

evgmik:
Indeed, the brig-read-ipfs:ten_enc:none-zip:{lz4,snappy} is very fishy, how it could be slower than the one with encryption. `fuse-read-ipfs` indeed of the charts. I tried `fuse-write-mem` and it is quite reasonable. Could it be that we are reading from the OS cache?

Here are my numbers:
```
$ brig debug iobench -b brig-read-ipfs -b fuse-read-ipfs  --size 100M

Time:          2021-02-09T21:32:17-05:00
CPU Name:      AMD Phenom(tm) II X4 965 Processor
Logical Cores: 4
Has AESNI:     no


ten:brig-read-ipfs
==================

brig-read-ipfs:ten_enc:none-zip:none          68.16    MB/s      100.00MB/1.467s   100.00%
brig-read-ipfs:ten_enc:none-zip:lz4           23.55    MB/s      100.00MB/4.247s    34.54%
brig-read-ipfs:ten_enc:none-zip:snappy        8.03     MB/s     100.00MB/12.455s    11.78%
brig-read-ipfs:ten_enc:aes256gcm-zip:none     40.68    MB/s      100.00MB/2.458s    59.69%
brig-read-ipfs:ten_enc:aes256gcm-zip:lz4      188.39   MB/s       100.00MB/531ms   276.39%
brig-read-ipfs:ten_enc:aes256gcm-zip:snappy   211.88   MB/s       100.00MB/472ms   310.86%
brig-read-ipfs:ten_enc:chacha20-zip:none      111.81   MB/s       100.00MB/894ms   164.04%
brig-read-ipfs:ten_enc:chacha20-zip:lz4       202.05   MB/s       100.00MB/495ms   296.44%
brig-read-ipfs:ten_enc:chacha20-zip:snappy    266.44   MB/s       100.00MB/375ms   390.90%

ten:fuse-read-ipfs
==================

fuse-read-ipfs:ten_enc:none-zip:none          1960.50  MB/s        100.00MB/51ms   100.00%
fuse-read-ipfs:ten_enc:none-zip:lz4           2005.99  MB/s        100.00MB/50ms   102.32%
fuse-read-ipfs:ten_enc:none-zip:snappy        2043.19  MB/s        100.00MB/49ms   104.22%
fuse-read-ipfs:ten_enc:aes256gcm-zip:none     2019.76  MB/s        100.00MB/50ms   103.02%
fuse-read-ipfs:ten_enc:aes256gcm-zip:lz4      1871.96  MB/s        100.00MB/53ms    95.48%
fuse-read-ipfs:ten_enc:aes256gcm-zip:snappy   1991.78  MB/s        100.00MB/50ms   101.60%
fuse-read-ipfs:ten_enc:chacha20-zip:none      1918.81  MB/s        100.00MB/52ms    97.87%
fuse-read-ipfs:ten_enc:chacha20-zip:lz4       1956.11  MB/s        100.00MB/51ms    99.78%
fuse-read-ipfs:ten_enc:chacha20-zip:snappy    1959.59  MB/s        100.00MB/51ms    99.95%
```

sahib:
> Could it be that we are reading from the OS cache?

Yep, seems like it. :+1: 

At least for FUSE this was the issue. [After opening the file to read in direct I/O mode](f8d9c5bbd2281ce95e6e73cebec05103d370bd2d) the numbers looks way more reasonable:

```
$ brig debug iobench -b fuse-read-mem --size 100M

Time:          2021-02-10T20:07:14+01:00
CPU Name:      Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
Logical Cores: 4
Has AESNI:     yes


ten:fuse-read-mem
=================

fuse-read-mem:ten_enc:none-zip:none           274.02   MB/s       100.00MB/365ms   100.00%
fuse-read-mem:ten_enc:none-zip:lz4            222.74   MB/s       100.00MB/449ms    81.28%
fuse-read-mem:ten_enc:none-zip:snappy         262.37   MB/s       100.00MB/381ms    95.75%
fuse-read-mem:ten_enc:aes256gcm-zip:none      167.81   MB/s       100.00MB/596ms    61.24%
fuse-read-mem:ten_enc:aes256gcm-zip:lz4       159.44   MB/s       100.00MB/627ms    58.19%
fuse-read-mem:ten_enc:aes256gcm-zip:snappy    259.04   MB/s       100.00MB/386ms    94.53%
fuse-read-mem:ten_enc:chacha20-zip:none       228.41   MB/s       100.00MB/438ms    83.35%
fuse-read-mem:ten_enc:chacha20-zip:lz4        221.77   MB/s       100.00MB/451ms    80.93%
fuse-read-mem:ten_enc:chacha20-zip:snappy     320.74   MB/s       100.00MB/312ms   117.05%
```

A note to something I think you mentioned already: We might run into problems when the fuse filesystem content gets changed without using fuse. The kernel only invalidates the page cache when writes happen to fuse, but it does not notice when we e.g. sync with another peer. Then there will be a short time where the OS still delivers the old data to the client, which is likely not a huge issue but still weird. Do you know if there is a way to invalidate the page cache for a fuse filesystem or even certain files? If yes, we should do this for 

EDIT: I know of ``sync; echo 3 > /proc/sys/vm/drop_caches``, but that clears all caches which is undesirable.

evgmik:
> A note to something I think you mentioned already: We might run into problems when the fuse filesystem content gets changed without using fuse. The kernel only invalidates the page cache when writes happen to fuse, but it does not notice when we e.g. sync with another peer. Then there will be a short time where the OS still delivers the old data to the client, which is likely not a huge issue but still weird. Do you know if there is a way to invalidate the page cache for a fuse filesystem or even certain files? If yes, we should do this for
> 
> EDIT: I know of `sync; echo 3 > /proc/sys/vm/drop_caches`, but that clears all caches which is undesirable.

Only `fuse.InvalidateNodeData(f)` comes to my mind as it shown at
https://github.com/bazil/fuse/blob/fb710f7dfd05053a3bc9516dd5a7a8f84ead8aab/examples/clockfs/clockfs.go#L160

But we would need a callback to fuse from main server to call invalidation. Which sound painful.

evgmik:
> brig debug iobench -b fuse-read-mem --size 100M

I am envy of your benchmark, factor of 2 or 3 faster then mine.

Though, I am still suspicious
```
$ brig debug iobench -b fuse-read-mem --size 100M

Time:          2021-02-10T21:54:37-05:00
CPU Name:      AMD Phenom(tm) II X4 965 Processor
Logical Cores: 4
Has AESNI:     no


ten:fuse-read-mem
=================

fuse-read-mem:ten_enc:none-zip:none              118.66 MB/s       100.00MB/843ms   100.00%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:none-zip:lz4                79.91 MB/s      100.00MB/1.251s    67.34%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:none-zip:snappy            127.99 MB/s       100.00MB/781ms   107.86%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:aes256gcm-zip:none          23.25 MB/s      100.00MB/4.301s    19.59%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:aes256gcm-zip:lz4           85.58 MB/s      100.00MB/1.169s    72.12%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:aes256gcm-zip:snappy        97.46 MB/s      100.00MB/1.026s    82.13%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:chacha20-zip:none           62.98 MB/s      100.00MB/1.588s    53.07%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:chacha20-zip:lz4            85.05 MB/s      100.00MB/1.176s    71.67%      4 allocs   100.00% zipped
fuse-read-mem:ten_enc:chacha20-zip:snappy        115.48 MB/s       100.00MB/866ms    97.32%      4 allocs   100.00% zipped
```

I do not think that I can get 100 MB/s from my spinning hard drive. Usually, I saturate it around 40 MB/s on a long sequential read.
If I do direct read/write with `dd` and `fuse` I see similar 40 MB/s.

sahib:
> I do not think that I can get 100 MB/s from my spinning hard drive. Usually, I saturate it around 40 MB/s on a long sequential read. If I do direct read/write with dd and fuse I see similar 40 MB/s.

Thing is, you don't use your hard drive here. All is done in ``/tmp`` which is probably a ``tmpfs`` for you. Also, dd is not comparable either, since we do in-memory operations here. We read from fuse and directly discard the data. ``dd`` has to open ``/dev/null`` and actually copy the data to it.

sahib:
Also if you envy my laptop, then you should look at the benchmark done [on more recent hardware](https://gist.github.com/serztle/e9f31ae7dc19670c380a7b7d528fb9ed) ([my laptop](https://gist.github.com/sahib/769fa893fa69e2bb20f2e5331615ab1a))

evgmik:
> Also if you envy my laptop, then you should look at the benchmark done [on more recent hardware](https://gist.github.com/serztle/e9f31ae7dc19670c380a7b7d528fb9ed) ([my laptop](https://gist.github.com/sahib/769fa893fa69e2bb20f2e5331615ab1a))

Wow. But this raises another concern how is it possible to get 12GB/s? It is faster then the clock speed. I have not look at new memory bus innovations, but I thought that you move about a byte per clock cycle. 

From other hand if you move 8 bytes (64 bits) than it make more sense, but I think memory transfer has nothing to do with 'bitness' of the CPU.

sahib:
> Wow. But this raises another concern how is it possible to get 12GB/s? It is faster then the clock speed. I have not look at new memory bus innovations, but I thought that you move about a byte per clock cycle.

That's basically the throughput of ``man 2 memcpy`` - okay, our numbers are probably still less than that, since we do a bit more than only ``memcpy``. It's how fast a userland process can move memory from A to B. A is the source (in the case just a 100M buffer with data) and B is the copy buffer passed to ``io.Copy()``.

> From other hand if you move 8 bytes (64 bits) than it make more sense, but I think memory transfer has nothing to do with 'bitness' of the CPU.

You can copy more than one byte per cycle. I'm also not sure how many, but copying a 64bit pointer on my system is an atomic and happens in one cycle. Also speculative/parallel execution could help here, but I'm not an expert regarding that. In any case, the theoretic limit of how fast memory can be moved is way higher than 12GB/s.

But anyways, I think the limiting factor here is less the CPU speed but more how fast your RAM bus can deliver the data. My laptop has DDR4 RAM with 2.4GHz. How we manage to get over that limit is beyond my knowledge, but the numbers we get do not seem unusual for ``memcpy``. Maybe @serztle can help? He ran the Ryzen benchmark above.

sahib:
The remaining problem of the fishy benchmark is now also resolved: 848adf9c344ae76617be315c42d00595533f0c27
I think I can merge this util now. It can be improved along the way.

-----

> Only fuse.InvalidateNodeData(f) comes to my mind as it shown at
https://github.com/bazil/fuse/blob/fb710f7dfd05053a3bc9516dd5a7a8f84ead8aab/examples/clockfs/clockfs.go#L160
> But we would need a callback to fuse from main server to call invalidation. Which sound painful.

Thanks, that looks promising. I don't think it will be too painful. FUSE should ask ``catfs`` when any file changed and if so, invalidate every file. Everything is changed over Stage anyways. I will write a ticket for this later, it's not too urgent though.

evgmik:
Small things.

1. Fuse related test, for example `fuse-write-mem` leave unmounted record like
```
brigfs on /tmp/brig-fuse-bench-039484116/mount type fuse.brig (rw,nosuid,nodev,relatime,user_id=1000,group_id=1000)
```
Which is a bit strange, since the benchmark process is finished and should have unmount everything by itself.

2. Same benchmark, generate quite a few files in the mount point. It is probably good idea to delete a test file before benchmark moves to next test of encoder/compressor. With small bench buffer size it is ok, but it keeps useful memory allocated.

sahib:
I could not resist to have some fun while investigating the high alloc counts showing in the output above, That led to some very noticeable improvements, especially in the compression layer:

https://gist.github.com/sahib/446284537b14afa65ab90ae4b107f537

The change are basically:

* Used another library for ``lz4``. The one we used before was not maintained anymore and rather slow too.
* Reduced allocations during compression and decompression. Now we allocate just once per stream. That could be optimized further with a ``sync.Pool``, but that would show only when handling many parallel streams. Notice the low alloc numbers in the link above.
* Added ``zstd`` support. I did not bother to add ``brotli`` since the only acceptable go library does not seem to support block based encoding.

Interesting observation: I added a ``mixed`` input type for the benchmark. ``zstd`` really seems to loose here. It performs well with high-compressible data and okay-ish with random data. So heuristic should only recommend ``zstd`` for things like text files and similar. While it yields the highest compression ratio of all, it seems too have rather flaky performance properties.

repo/hints/hints.go
```diff
@@ -48,14 +51,16 @@ var (
...
 		CompressionGuess:  compress.AlgoUnknown,
 	}
 
 	compressionSortMap = map[CompressionHint]int{
 		CompressionNone:   0,
 		CompressionLZ4:    1,
 		CompressionSnappy: 2,
-		CompressionGuess:  3,
+		CompressionZstd:   3,
+		CompressionGuess:  4,
```
evgmik: I would move ComressionGuess to `1`. It is likely we will add/remove algorithms, moving guess number
will be inconvinient

evgmik (approved):

catfs/mio/compress/algorithm.go
```diff
@@ -50,74 +55,135 @@ func (at AlgorithmType) String() string {
...
-		"snappy": AlgoSnappy,
-		"lz4":    AlgoLZ4,
+		AlgoZstd:   "zstd",
 	}
 )
 
 // AlgoSnappy
-func (a snappyAlgo) Encode(src []byte) ([]byte, error) {
-	return snappy.Encode(nil, src), nil
+func (a snappyAlgo) Encode(dst, src []byte) ([]byte, error) {
```
evgmik, line 142: It is purely to satisfy my curiosity: but don't we already have `dst` with all the data. What is the need to return the byte range? It clearly crucial, but I do not see why.

evgmik (approved):

catfs/mio/compress/algorithm.go
sahib: The ``dst`` is the complete buffer, returned is a sub slice of ``dst`` where ``len()`` indicates the length of the compressed (or decompressed) data. This is always true though: ``len(returned) <= len(dst)``.

repo/hints/hints.go
sahib: It's purely for sorting and that's mostly for visuals, but sure I can change it.

evgmik:
> I could not resist to have some fun while investigating the high alloc counts showing the output above, That led to some very noticeable improvements, especially in the compression layer:

Nice one! This is huge speed up.

> Interesting observation: I added a `mixed` input type for the benchmark. `zstd` really seems to loose here. It performs well with high-compressible data and okay-ish with random data. So heuristic should only recommend `zstd` for things like text files and things like those. While it yields the highest compression ratio of all, it seems too have rather flaky performance properties.

I was expecting a more stable performance from hyped `zstd`. But I guess everyone can fail on a particular subset of data. It was interesting to see how speed of `zstd` drops on the mixed dataset.

sahib:
> Fuse related test, for example fuse-write-mem leave unmounted record like

Teardown code was a little buggy, should be better now.


>  Same benchmark, generate quite a few files in the mount point. It is probably good idea to delete a test file before benchmark moves to next test of encoder/compressor. With small bench buffer size it is ok, but it keeps useful memory allocated.

True, but that's a bit harder to fix since I would need to set-up the benchmark on each run which takes quite long. If it happens I lowered the samples and buffer size. If it becomes too annoying we can come back to it later.

PR #94, opened 2021-02-10
Fix/fixing regression in test suite

evgmik:
Some test suit fixes, and test regression fix after adding Hints interface

### DIFF
--- a/client/fs_test.go
+++ b/client/fs_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/sahib/brig/repo"
 	"github.com/sahib/brig/server"
+	"github.com/sahib/brig/repo/hints"
 	colorLog "github.com/sahib/brig/util/log"
 	"github.com/sahib/brig/util/testutil"
 	log "github.com/sirupsen/logrus"
@@ -491,8 +492,9 @@ func TestHints(t *testing.T) {
 		info, err := ctl.Stat(path)
 		require.NoError(t, err)
 
-		require.Equal(t, "guess", info.Hint.CompressionAlgo)
-		require.Equal(t, "aes256gcm", info.Hint.EncryptionAlgo)
+		defHints := hints.Default()
+		require.Equal(t, string(defHints.CompressionAlgo), info.Hint.CompressionAlgo)
+		require.Equal(t, string(defHints.EncryptionAlgo), info.Hint.EncryptionAlgo)
 		require.Equal(t, false, info.IsRaw)
 
 		none := "none"
--- a/client/fs_test.go
+++ b/client/fs_test.go
@@ -38,7 +38,18 @@ func withDaemon(t *testing.T, name string, fn func(ctl *Client)) {
 	repoPath, err := ioutil.TempDir("", "brig-client-repo")
 	require.Nil(t, err)
 
-	defer os.RemoveAll(repoPath)
+	defer func() {
+		// Somehow there is race condition between 
+		// srv.Close() from the defer at the very end 
+		// os.RemoveAll(repoPath).
+		// Theoretically, `go` should have closed server
+		// but in practice I see that repoPath is removed
+		// before server had a chance to close the DB
+		// and I see complains in log about DB.Close
+		// I introduce this time delay as a crude hack
+		time.Sleep(100 * time.Millisecond)
+		os.RemoveAll(repoPath)
+	}()
 
 	daemonURL := "unix:" + filepath.Join(repoPath, "brig.socket")
 	err = repo.Init(repo.InitOptions{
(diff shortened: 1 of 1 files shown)


client/fs_test.go
```diff
@@ -38,7 +39,18 @@ func withDaemon(t *testing.T, name string, fn func(ctl *Client)) {
...
+	defer func() {
+		// Somehow there is race condition between 
+		// srv.Close() from the defer at the very end 
+		// os.RemoveAll(repoPath).
+		// Theoretically, `go` should have closed server
+		// but in practice I see that repoPath is removed
+		// before server had a chance to close the DB
+		// and I see complains in log about DB.Close
+		// I introduce this time delay as a crude hack
+		time.Sleep(100 * time.Millisecond)
```
sahib, line 51: Hmm, not so convinced by this reasoning. Go defers are put on a stack and run in inverse order when leaving the function. Therefore the server close should come first... maybe something else deletes the directory before the close comes? This defer here should be executed last.  It could be also that a go routine in the server still runs when Close exists. Then this should be fixed instead.

sahib (changes requested):
One test failed (might be related):

```
=== RUN   TestClientCompleteFetchAllowed
time="2021-02-10T04:44:12Z" level=warning msg="Failed to retrieve own pubkey: open /tmp/brig-ctl-test500264232/keyring/alice/key.pub: no such file or directory"
time="2021-02-10T04:44:13Z" level=info msg="starting connection with addr `bob@45575`"
time="2021-02-10T04:44:13Z" level=info msg="pinger »bob@45575« responding"
time="2021-02-10T04:44:13Z" level=info msg="starting connection with addr `alice@40579`"
time="2021-02-10T04:44:13Z" level=info msg="pinger »alice@40579« responding"
time="2021-02-10T04:44:13Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:45575->127.0.0.1:50672: write: broken pipe"
time="2021-02-10T04:44:13Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:40579->127.0.0.1:44792: write: broken pipe"
time="2021-02-10T04:44:14Z" level=info msg="Will not accept new connections now"
time="2021-02-10T04:44:14Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:45575->127.0.0.1:50688: write: broken pipe"
time="2021-02-10T04:44:14Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:40579->127.0.0.1:44810: write: broken pipe"
time="2021-02-10T04:44:14Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:40579->127.0.0.1:44826: write: broken pipe"
time="2021-02-10T04:44:15Z" level=warning msg="failed to authenticate connection: write tcp 127.0.0.1:40579->127.0.0.1:44842: write: broken pipe"
time="2021-02-10T04:44:15Z" level=info msg="Will not accept new connections now"
time="2021-02-10T04:44:15Z" level=warning msg="Failed to retrieve own pubkey: open /tmp/brig-ctl-test710451505/keyring/alice/key.pub: no such file or directory"

    client_test.go:37: 
        	Error Trace:	client_test.go:37
        	            				client_test.go:89
        	            				client_test.go:112
        	            				client_test.go:255
        	Error:      	Expected nil, but got: &os.PathError{Op:"unlinkat", Path:"/tmp/brig-ctl-test710451505", Err:0x27}
        	Test:       	TestClientCompleteFetchAllowed
--- FAIL: TestClientCompleteFetchAllowed (3.54s)
```

client/fs_test.go
evgmik: Yes, I agree that *theoretically* defers go in FILO order. But maybe badger Close() has some wierd logic and delayed execution.
Removing this Sleep() brings warning messages in the Test log.

The test itself is not failing but it is log clearly shows warning about unable to close DB. 

The test which you pointed out, passes for me. But I am quite sure that the issue is similar. The strange part we have not changed anything in DB related code. It is really a combination of HDD activity and the system load.

evgmik: Also the race condition would explain random fails with Travis test runs.

sahib: Well, I would prefer to have the original race condition fixed instead of the workaround, but the fix is acceptable for the time being. :smirk: Approved.

sahib (approved):

client/fs_test.go
evgmik: > Well, I would prefer to have the original race condition fixed instead of the workaround, but the fix is acceptable for the time being. smirk Approved.

I did a bit of digging, I suspect it is because `server.Close()` uses channels to communicate the quit/close. Thus we exit function before the intended event `db.Close()`. This is where my expertise unfortunately ends, I guess we need to wait for acknowledgement from the go routine.

Issue #92, opened 2021-02-07
New database for brig


evgmik:
For the `brig` database we are currently using `badger` v1.6.2 and there are thoughts to move to a more modern verstion 3.2011.1 (at the time of writing). But first we should think is `badger` is the right DB.

`badger` use different storage for key and value, which allows to do manipulation with keys very fast. `badger` is the fastest according to their benchmark. But this kind of DB benefit the most when keys are short and values are large. This is technically not the case for the `brig` structure, our keys are hashes which are around 50 bytes, and keys are quite short or comparable. On one test instance the raw (absolute minimum without compression) size of the keys is 9151 B and values 16568. So a typical value twice larger than the key. More importantly, the values are small 1-2kB. So `badger` which was designed for big data brings us no benefits. In fact we have huge data write size amplification due to `badger` internal structure. In above case you would need 9151+16568 = 25719 of raw storage, yet `badger v1.6.2 db` takes 517kB (all it stores about 100 commits of the same filename with a different content). So we see write amplification of 20 times. The raw size could be even smaller, if we use more compact values encoding. Also we see, that a simple `brig touch newfile` takes 3-4kB of DB storage.

Sure, we all have GB of HDD but there is clearly an inefficiency.

evgmik:
(the comment was updated 2021-02-07 with correct numbers)

Storage efficiency of badger db for different versions. 

I modified the script posted https://github.com/dgraph-io/badger/issues/686#issue-399905584

The script generates 100,000 key:value pairs with each key and each value being 10bytes.
```
package main

import (
	"github.com/dgraph-io/badger"
	"log"
	"fmt"
)

func main() {
	opts := badger.DefaultOptions("/tmp/badger")
	db, err := badger.Open(opts)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	wb := db.NewWriteBatch()
	defer wb.Cancel()

	for i := 0; i < 100000; i++ {
		key := fmt.Sprintf("key_%06d", i)
		val := fmt.Sprintf("val_%06d", i)
		err := wb.Set([]byte(key), []byte(val)) // Will create txns as needed.
		if err != nil {
			log.Fatal(err)
		}
	}
	wb.Flush()
	fmt.Println("Bulk upload done")
}
```

The script generates 10,000 key:value pairs with each key and each value being 10bytes. So the minimal storage space is  2MB = 2,000,000 bytes.

[...]

PR #96, opened 2021-02-12
Feature/badger db v3

evgmik:
Addresses #92

`badger v3.2011.1` made a few old options obsolete so they are removed.
I have adjusted the value log file to 10MB (from default 2GB), and memory table size to 10MB from default 128MB. If we do our job right we do not need that much anyway.

Another reason for this changes. `badger` allocate above values on hard drive, if user does not use modern FS which allow sparse files (with zeros). He or she would see 2GB of space disappeared before any operation with `brig` starts.
On modern FS, one would see *actual* allocation. Still 2GB file size looks scary to me.

So far, I see no negative effects from the badger version upgrade.

To to see compression improvements, the DB need to be garbage collected which can be done only at db.Close().


evgmik:
Looks like we ready to merge with develop.

I decided to skip GC for `gateway/db/db.go` it is rarely touched and is not poluted with obsolete records.

By the way, gateway db has a lot of interfaces which seems to be copy pasted from `catfs/db/database_badger.go`. Maybe there is a way to unite them, then they can use the same GC routine.

sahib:
> I decided to skip GC for gateway/db/db.go it is rarely touched and is not poluted with obsolete records.

Fine to me.

> By the way, gateway db has a lot of interfaces which seems to be copy pasted from catfs/db/database_badger.go. Maybe there is a way to unite them, then they can use the same GC routine.

Right. Might use badger also for as persistent cache for overlay caching. In that process I might add a generic KV-util, if I think the abstraction is worth the effort.

catfs/db/database_badger.go
```diff
@@ -56,17 +55,95 @@ func NewBadgerDatabase(path string) (*BadgerDatabase, error) {
...
+	defer bdb.mu.Unlock()
+	defer func() {
+		bdb.gcNeededAt = time.Now()
+		bdb.needsGC = false
+	}()
+	var errGC error
+	var beforeGC uint64 = 0
+	var total uint64 = 0
+	var err error
+	log.Infof("Performin GC for badger DB in %s", opts.Dir)
```
sahib: typo: "Performing".

catfs/db/database_badger.go
```diff
@@ -56,17 +55,95 @@ func NewBadgerDatabase(path string) (*BadgerDatabase, error) {
...
+			log.Infof("At this iteration DB size in %s decreased by %s", opts.Dir, humanize.Bytes(total-newTotal))
+		}
+		total = newTotal
+
+		if errGC = bdb.db.RunValueLogGC(0.5); errGC != nil {
+			if errGC == badger.ErrNoRewrite {
+				log.Debugf("badger DB in %s has nothing for GC",opts.Dir)
+			}
+			continue
+		}
```
sahib, line 127: What's the purpose of still running ``RunValueLogGC(0.5)`` after Close? Is it essential to call?

catfs/db/database_badger.go
```diff
@@ -56,17 +55,95 @@ func NewBadgerDatabase(path string) (*BadgerDatabase, error) {
...
+			return err
+		}
+		bdb.db, err = badger.Open(opts)
+		if err != nil {
+			log.Fatalf("Could not reopen DB in %s", opts.Dir)
+			return err
+		}
+		// Badger people use os.FileInfo to get total size of lsm and vlog files
+		// This is not actual but allocated of hdd size
+		lsm, vlog := bdb.db.Size() // values could be stale even after Close/Open cycle, badger updates it once per minute
```
sahib: Do we really need all this complicated logic if it's only for outputting info messages?

catfs/db/database_badger.go
```diff
@@ -323,6 +400,10 @@ func (db *BadgerDatabase) Flush() error {
 
 	db.txn = nil
 	db.haveWrites = false
+	if time.Now().After(db.gcNeededAt) && !db.needsGC {
+		db.gcNeededAt = time.Now().Add(120 * time.Second) // badger updates stats once per minute, we give ample time for stats to settle
```
sahib: Just a style thing: Please put the comments above the line they're referring to to avoid extra long lines.

sahib (changes requested):

catfs/db/database_badger.go
evgmik: fixed, but there is a new patch which makes it obsolete.

catfs/db/database_badger.go
evgmik: According to the Note in https://github.com/dgraph-io/badger/issues/767#issuecomment-485713746
the compactors,  which collect statistics for GC, are run only on DB `Close()` at least on small size DB.
So `Close` is needed for our DB size.

But your comment made me do more digging. So some things about `badger` DB GC collector was wrongly reversed engineered by me: `RunValueLogGC()` actually compacts on the fly, no need to `Close()` to shrink the DB. The problem that `RunValueLogGC` does not collect stats by itself, but uses prepared by compactors.
Which leads for necessity of one `Close()` to make sure that compactors are forced. Apparently, even `Close`
does not force it and need about 1 minute delay between DB modification and `Close` for compactors to kick in.

So new 3cc667c6 patch get rid of cycled Close/Open (it does it only once per `runGC` call), also I get rid
of the timer since the oldest DB flush and the modification flag.

If only there were reliable way to check if DB was modified, then we would not do unnecessary `Close()`. Though I noticed that `filesystem GC` constantly messing with the metadata DB even on idle `brig`.

catfs/db/database_badger.go
```diff
@@ -56,17 +51,60 @@ func NewBadgerDatabase(path string) (*BadgerDatabase, error) {
...
+	// But unfortunately for our typical size (even as high as 150 MB)
+	// compaction, even with Flatten(), does not kick in.
+	// The only way to truly force compaction (to update stats for GC) is to close DB
+	// see Note in https://github.com/dgraph-io/badger/issues/767#issuecomment-485713746
+	// After Close() the GC on a next run will have updated statistic
+	// Actually even Close() does not guaranteed success, it requires more than a minute
+	// to update stats after DB was modified. But eventually GC stats will be ready.
+	err := bdb.db.Close()
+	if err != nil {
+		log.Fatalf("Could not close DB in %s", opts.Dir)
```
sahib, line 132: I generally see calling ``log.Fatalf`` as bad practice, especially if it's not in quick & dirty programs and especially in library code (I have a story to tell here). I'm not fully sure what the proper error handling here is, but we could enter a retry loop when fail to close or open the database.

catfs/db/database_badger.go
sahib: Thanks for the explanation. Code looks better to me now.

One simple to gather statistic we could display would be to display how long the GC run took. I'm a little worried that it blocks all of brigs operation for considerable time, even if there weren't many writes in that time period.

sahib (commented):
There are two optional subjects that we should discuss before merging. Otherwise looks good to me :+1:

catfs/db/database_badger.go
evgmik: I tried to mitigate with 91d48265. Error in `Close` is not fatal, we can try later. If we cannot open, then we do ten attempts with timeouts and then Fail. We cannot run without DB. Also, fixed Close which did not check for an active transaction. GC could be done or running system, but Close definitely not.

I would love to hear this story :) I was always thinking that if something going wrong the program should faill with error message so user would send an angry bug report and we fix the problem. If we are running industrial process, then the story is different.

evgmik:
I have added GC elapsed time the worst time is about 170mS after big cleanup. Typical time is about 15mS with Close/Open of DB before patch f7ae738.

With patch f7ae738 GC is very quick, since it does not do unnecessary (Close/Open) work on GC if stats are known to be not refreshed

catfs/db/database_badger.go
sahib: > Error in Close is not fatal

Is it? What if it was partially closed and we're left with a zombie db conn?

> I would love to hear this story :)  If we are running industrial process, then the story is different.

Story is: We once used a [gpio library](https://github.com/brian-armstrong/gpio/pull/16/files) to check some pins on our device. Sadly I didn't look into the library implementation, so that fault is on me, but their error handling was "log & exit". Which is a bad strategy as a library. I noticed this first when the program didn't first run on my work machine (obvious cause was that the gpio was not available there), but at some point we made the grave mistake to update the Go version to 1.14 and all hell broke loose. All services reading this pin suddenly crashed randomly after the update. And why was that? Because since 1.14 ``select()`` might return ``EINTR`` (rightly so) when it's execution was interrupted by the scheduler, which might happen at any point. The gpio library was not prepared for this and just jumped off the cliff by calling ``os.Exit(1)``. Fix was easy, but what we can learn from this:

* Never ever ``os.Exit()`` in library code.
* Check the quality of the library you're using.
* Error handling is hard.

> I was always thinking that if something going wrong the program should faill with error message so user would send an angry bug report and we fix the problem.

That's fine for interactive programs. Long running programs (and brig daemon counts as one) should do this only as last resort since the error gets buried in some logs (in the best case). But not being able to open the main database is one such case, so https://github.com/sahib/brig/commit/91d482653845748445bcb0e91335de793bc09acc is fine with me.

sahib (approved):

evgmik:
You are right:  `for retries := 0; retries < 10; retries++ { ... }` is much cleaner. Fixed in 5e5821c1.
In the same patch retrying `db.Close`. You are right yet again, it could be left in the half broken state after partial `Close`

catfs/db/database_badger.go
evgmik: I see now why it was a painful experience. Indeed `fail` is super special and should be used as a last line of defense.

catfs/db/database_badger.go
evgmik: done

PR #97, opened 2021-02-13
fix: the file system `badger db` is now closed on Quit

evgmik:
Somehow, we never closed FS DB on quit or interrupt. Miraculously, `brig`
worked on restart as if nothing happened. But the DB was left in a dirty
state, with DB `LOCK` file left and some other transient tables.


sahib:
Good find! :+1:

sahib (approved):

PR #99, opened 2021-02-14
cmd: allow specifiyng a ipfs-url instead of a ipfs-path (--ipfs-path-…

### LINKED ISSUE #32: Allow for remote IPFS APIs
**Is your feature request related to a problem? Please describe.**
Since `brig` is now using the HTTP IPFS API, one should be able to connect to any IPFS daemon, not just the local one. This was asked about by @postables in the chatroom. 

**Describe the solution you'd like**
Instead of using the `--ipfs-path` flag, there should be an `--ipfs-api` flag and an `--ipfs-port` flag. The former will allow for specifying an address to a node such as `11.22.33.44/example`, or `example.com`. The latter option can specify a port to use, and would default to `5001`, the IPFS default. If `--ipfs-api` isn't specified `127.0.0.1` can be used.

**Describe alternatives you've considered**
You could connect a remote port to your local device using `ssh` as a hack for the time being.

sahib:
This closes #32. We now allow the following form of ``brig init``:

```bash
brig init --repo . --ipfs-path-or-url /ip4/127.0.0.1/tcp/5002 some-name
```

This will connect to an existing IPFS daemon at the specified mutliaddr.

### DIFF
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -83,7 +83,7 @@ var helpTexts = map[string]helpEntry{
 			},
 			cli.StringFlag{
 				Name:  "ipfs-path-or-multiaddr",
-				Usage: "Specify an explicit path or URL to an IPFS repository. Useful if you have more than one.",
+				Usage: "Specify a path to an ipfs repo (/tmp/ipfs-repo) or a multiaddr to a running ipfs daemon (/ip4/127.0.0.1/tcp/5002)",
 				Value: "",
 			},
 			cli.BoolFlag{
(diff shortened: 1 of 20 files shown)


cmd/help.go
```diff
@@ -82,8 +82,8 @@ var helpTexts = map[string]helpEntry{
 				Usage: "Do not display the super pretty logo on init.",
 			},
 			cli.StringFlag{
-				Name:  "ipfs-path,P",
-				Usage: "Specify an explicit path to an IPFS repository. Useful if you have more than one.",
+				Name:  "ipfs-path-or-url",
```
evgmik: It would be nice to show an example of expected form of url (and even path).

evgmik (approved):

cmd/help.go
sahib: Done. Also renamed to ``--ipfs-path-or-multiaddr`` since it wasn't a URL, but a multiaddr.

Issue #100, opened 2021-02-20
fuse layer creates backend pin twice


evgmik:
Steps to reproduce.
1. Mount `brig` via fuse
2. Copy any file to mount
Observe in logs:
```
fuse/util.go:18: errorize: dir-lookup: No such file: No such file or directory: /ztes2
20.02.2021/14:48:41 ⚙ catfs/mio/stream.go:138: guessed 'unknown' compression for file /ztes2
20.02.2021/14:48:41 ⚙ catfs/core/coreutils.go:470: adding /ztes2 (QmSG3yLbF6VfimqSiXoB8kSL8PkmC3w7X2zcaUsqFV2tKG)
20.02.2021/14:48:41 ⚙ fuse/file.go:37: fuse-file-attr: /ztes2
20.02.2021/14:48:41 ⚙ fuse/handle.go:75: fuse-Write: /ztes2 (off: 0 size: 4897)
20.02.2021/14:48:41 ⚙ fuse/file.go:119: fuse-file-setattr: request Setattr [ID=0x1a Node=0x4 Uid=1000 Gid=1000 Pid=24328] mtime=2021-02-20 14:48:41.545509997 -0500 EST handle=0x1
20.02.2021/14:48:41 ⚙ fuse/file.go:37: fuse-file-attr: /ztes2
20.02.2021/14:48:41 ⚐ fuse/handle.go:92: fuse: Write time 1.042131ms for 4897 bytes
20.02.2021/14:48:41 ⚙ fuse/handle.go:119: fuse-flush: /ztes2
20.02.2021/14:48:41 ⚙ catfs/mio/stream.go:138: guessed 'lz4' compression for file /ztes2
20.02.2021/14:48:41 ⚐ File exists; modifying. [file=/ztes2]
20.02.2021/14:48:41 ⚙ catfs/core/coreutils.go:470: adding /ztes2 (QmTshsWjvZG5x8ZLTsVSnVHYvbZwDP8Kuz6MBKJN9BdTou)
20.02.2021/14:48:41 ⚐ fuse/handle.go:129: fuse: Flashed `/ztes2` in 69.900689ms
20.02.2021/14:48:41 ⚙ fuse/handle.go:139: fuse-release: /ztes2
```

Note that `catfs/core/coreutils.go:470: adding /ztes2 (Qm...)` happens twice with different hash.

The first hash seems to always have size of 36 bytes.

evgmik:
[...]

PR #103, opened 2021-02-28
Fix/make touch zero size identical

evgmik:
addresses #100

### DIFF
--- a/catfs/fs.go
+++ b/catfs/fs.go
@@ -884,8 +884,11 @@ func (fs *FS) Touch(path string) error {
 	// We may not call Stage() with a lock.
 	fs.mu.Unlock()
 
-	// Nothing or a ghost there, stage an empty file.
-	return fs.Stage(prefixSlash(path), bytes.NewReader([]byte{}))
+	// Nothing or a ghost there, stage an empty file
+	// 0 sized (newly touched) files should have the same key 
+	// to point to the same backend file
+	key := make([]byte, 32)
+	return fs.stageWithKey(prefixSlash(path), bytes.NewReader([]byte{}), key)
 }
 
 // Truncate cuts of the output of the file at `path` to `size`.
@@ -982,7 +985,7 @@ func (fs *FS) preStageKeyGen(path string) ([]byte, error) {
 		return nil, err
 	}
 
-	if oldFile != nil {
+	if oldFile != nil && oldFile.Size() != 0{
 		return oldFile.Key(), nil
 	}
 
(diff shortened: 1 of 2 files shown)


catfs/fs.go
```diff
@@ -32,9 +32,14 @@ import (
 )
 
 const (
-	abiVersion = 1
+	abiVersion                 = 1
+	defaultEncryptionKeyLength = 32
 )
 
+func defaultEncryptionKey() []byte {
```
sahib: The name is maybe a bit misleading: It's not the default, nor the initial one. It's just the one for empty files.
Mentioning, because it would ring my alarms when I'd read this in a foreign code base.

catfs/fs.go
```diff
@@ -32,9 +32,14 @@ import (
 )
 
 const (
-	abiVersion = 1
+	abiVersion                 = 1
+	defaultEncryptionKeyLength = 32
```
sahib, line 36: Just a musing: It's a pure coincidence that key length is always 32 bytes, since both algorithms we support are 256bit. If we ever change that we have to make sure that we support different key sizes and make the logic depend on the chosen algorithm.

Nothing to do with this PR, just mentioning because I was reminded of that.

catfs/fs_test.go
```diff
@@ -603,28 +610,47 @@ func TestTouch(t *testing.T) {
...
-		require.Nil(t, err)
+		require.NoError(t, err)
 
-		require.Nil(t, fs.Stage("/x", bytes.NewReader([]byte{1, 2, 3})))
+		// Double Check that the empty file has defaultEncryptionKey
+		require.Equal(t, oldInfo.Key, defaultEncryptionKey())
 
-		require.Nil(t, fs.Touch("/x"))
+		// Check that two empty files have same backend hash
+		require.Equal(t, oldInfo.BackendHash, yInfo.BackendHash)
```
sahib, line 628: :+1:

sahib (approved):
Looks good to me. :+1:

catfs/fs.go
evgmik: Indeed, I was not clear enough. In 0776b1e8 I changed
`defaultEncryptionKey` to  `emptyFileEncryptionKey`

catfs/fs.go
evgmik: This one is a bit harder to correct. The definition of this constant might
belong to encryption stream code. Since you don't object, I leave it as is for now.

PR #104, opened 2021-03-07
Feature/persistent overlay

sahib:
Sooo... this took me way longer than anticipated, but here we are. I implemented a new pagecache that replaces the confusing ``catfs/mio/overlay`` with something better documented but also probably even more confusing.  I tried to explain most things with comments, i.e. in ``catfs/mio/pagecache/docs.go``.

Some comparison from the old system to the new one might help:

Old ``catfs/mio/overlay``:

* Reads that were not sequential caused a Seek() on the stream. The cause was that fuse did not know if the data was cached.
* Writes were are stored in memory, there was no size limit. Copying big files will use as much memory as the file is big.
* Every write caused on allocation in the same size of the write. Many small writes would cause many allocations and
  many copies, since overlapping (each with own allocation) writes
  were combined.

New ``catfs/mio/pagecache``:

* Reads only call Seek() if the data is not cached and the stream offset is not already where we want to be.
* Writes are stored until a limit in memory and are swapped to disk beyond it. While this is slower it is certainly better than bringing down the whole computer. I might add support to snappy-compress pages on disk and even memory.
* Memory is allocated in 64k-pages. The stream is divided in such pages and writes will be copied onto the affected pages. This uses a bit more memory (e.g. for files < 64k) but does few big allocations instead of many small ones. Also, the memory should not be actually used (i.e. virtual memory vs residual) until we actually copy things to it.

Some stuff from my TODO-list:

* Well, actually try things out. Had only a quick try and wrote quite some unit tests. So far it worked, but I didn't bother yet to add several big files (one 2G worked, just had the repo under ``/tmp`` which kinda defeats that...)
* Investigate in snappy compression for l1 and l2 cache.  Might be a great option if you want to trade some CPU.
* Add some more tests, especially one that randomly reads from the stream. I'm pretty sure the old implementation had some subtle bugs and I found some while writing a random-write test.
* Verify the assumption that memory is not actually used until we copy stuff to it (true in C, never checked in Go).
* Minor cleanups of the logic. I think I got confused in a few places and got over-cautious. But seriously that code really is the mother of one-off errors.


evgmik (approved):

evgmik (approved):

evgmik:
Wow. This looks quite complicated yet simple. Nothing bad jumped into my eye, though it said to see that other cache libraries are no good to use.

evgmik (approved):

sahib:
I think I'm finally done with this feature. Tested it a bit and it seems to work nicely. Did so by copying a big video to the mount and seeking around it. Working on a image in GIMP also seemed to work good. Pages over 1G were swapped to the disk cache, as expected.

I did not test performance yet though. My expectation is that it's slightly faster on write-heavy cases due to the reduced amount of allocations. I do not expect big changes though. Can be tackled at another point.

evgmik (approved):

evgmik:
If you could wait couple days with merge, I will also run couple in situ test. Maybe something unexpected will pop up. Though, the code look good and I do not expect troubles.

evgmik:
I tried new changes. Everything looks smooth. The only concern is that the log shows quite a few messages like
```
13.03.2021/18:06:47 ⚠ catfs/mio/pagecache/mdcache/l2.go:99: page l2: failed to delete /tmp/ali/pages/ali/00000020-0000050c
13.03.2021/18:06:47 ⚠ catfs/mio/pagecache/mdcache/l2.go:99: page l2: failed to delete /tmp/ali/pages/ali/00000020-0000050d
13.03.2021/18:06:47 ⚠ catfs/mio/pagecache/mdcache/l2.go:99: page l2: failed to delete /tmp/ali/pages/ali/00000020-0000050e
13.03.2021/18:06:47 ⚠ catfs/mio/pagecache/mdcache/l2.go:99: page l2: failed to delete /tmp/ali/pages/ali/00000020-0000050f
```

not sure if they of any importance

sahib:
> I tried new changes. Everything looks smooth. 

Cool!

> The only concern is that the log shows quite a few messages like [...]

Not a problem. Those just indicate "double" removals or things that were not in l2, but in l1. I removed the log.

PR #116, opened 2022-01-20
Readme: Update development build instructions

tennox:
For my issue reports I was asked if the error is still happening in develop binary, so I tried to figure out how to build it (as the instructions were outdated).

I managed to compile using the given instructions, so I guess this PR could be merged, but I'm not sure as the built binary is giving me trouble:

### brig path not detected
I need to set BRIG_PATH environment variable, otherwise the daemon wont launch, as it only checks for config.yml in current working dir.
I guess there's a [TODO related](https://github.com/sahib/brig/blob/6b7eccf8fcbd907fc759f8ca8aa814df8499e2ed/cmd/util.go#L81) ?

### brig repo of latest stable cannot be used
I get error of missing `immutable.yml` - I guess there is no migration process yet?
So I delete repo and `init` again, but...

### ipfs dir is not detected or set
It again tries to search for IPFS files (`config` dir I think) in current working dir, but I found I could set it manually in config.yml:
`ipfs_path_or_url: "/home/manu/.ipfs"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,24 +44,25 @@ Also take a look [at the documentation](http://brig.readthedocs.io/en/latest/ind
 
 ## Installation
 
-You can download the latest script with the following oneliner:
+You can use the installer script with the following oneliner:
 
 ```bash
 # Before you execute this, ask yourself if you trust me.
 $ bash <(curl -s https://raw.githubusercontent.com/sahib/brig/master/scripts/install.sh)
 ```
 
-Alternatively, you can simply grab the latest binary from the [release tab](https://github.com/sahib/brig/releases).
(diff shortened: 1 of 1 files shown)


sahib:
Hello @tennox,

thanks for your report. Your difficulties seem to arise due to some out dated documentation, since a lot things changed in the ``develop`` branch. Some things are not ready for a merge yet, which is a rather unfortunate state. Sadly, I currently have no time to fix all of this (and the other tickets you mentioned) due to time restrictions on my side. Once (and if) I come back to it, I will check this and work towards a merge. Then also this PR will be merged.

